PR #27, opened 2018-05-27
Added KeyboardShortcut class for easy way of capturing keybinds with modifiers

ManlyMarco:
Made this while working on the mod config screen/manager. Should be useful for modders regardless if they want their keybinds to be editable by other people. Including this in BepInEx will also allow people to avoid referencing the config manager. 

Comment from the class:
A keyboard shortcut that can be used in Update method to check if user presses a key combo.
Modifier keys are handled properly - if your keybind is Space, and Space+Shift is pressed, your keybind will not fire.
Use SavedKeyboardShortcut version to automatically save any changes made to this keybind.
How to use: Use IsDown instead of the Imput.GetKeyDown in the Update loop.

### DIFF
--- a/BepInEx/KeyboardShortcut.cs
+++ b/BepInEx/KeyboardShortcut.cs
@@ -1,0 +1,162 @@
+﻿using System;
+using System.ComponentModel;
+using System.Linq;
+using UnityEngine;
+
+namespace BepInEx
+{
+    /// <summary>
+    /// A keyboard shortcut that can be used in Update method to check if user presses a key combo.
+    /// Modifier keys are handled properly - if your keybind is Space, and Space+Shift is pressed, your keybind will not fire.
+    /// Use SavedKeyboardShortcut version to automatically save any changes made to this keybind.
+    ///
+    /// How to use: Use IsDown instead of the Imput.GetKeyDown in the Update loop.
+    /// </summary>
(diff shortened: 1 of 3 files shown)


BepInEx/KeyboardShortcut.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+            return Input.GetKeyDown(Key) && ModifierKeyTest();
+        }
+
+        /// <summary>
+        /// Check if the main key is currently held down (Input.GetKeyDown), and specified modifier keys are all pressed
+        /// </summary>
+        public bool IsPressed()
+        {
+            return Input.GetKeyDown(Key) && ModifierKeyTest();
```
denikson: What's the difference from `IsDown`? Just the name? Or should this be instead the following?

```csharp
return Input.GetKey(Key) && ModifierKeyTest();
```

More info: [Input.GetKey](https://docs.unity3d.com/ScriptReference/Input.GetKey.html).

ManlyMarco: You are correct, it was supposed to be Input.GetKey

ManlyMarco:
Fixed the issue

ManlyMarco:
Decided to put these into BepisPlugins.

PR #88, opened 2019-08-12
Overhaul config system

ManlyMarco:


BepInEx/Contract/PluginInfo.cs
```diff
@@ -3,7 +3,7 @@
 using System.Linq;
 using BepInEx.Bootstrap;
 
-namespace BepInEx.Contract
+namespace BepInEx
```
denikson: Any reason for swapping the namespace?

BepInEx.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28922.388
+# Visual Studio 15
```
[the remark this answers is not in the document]
ManlyMarco: VS automatically did it, I can roll back if needed

BepInEx/Configuration/AcceptableValueList.cs
```diff
@@ -0,0 +1,46 @@
...
+			if (acceptableValues == null) throw new ArgumentNullException(nameof(acceptableValues));
+			if (acceptableValues.Length == 0) throw new ArgumentException("At least one acceptable value is needed", nameof(acceptableValues));
+
+			_acceptableValues = acceptableValues;
+		}
+
+		/// <inheritdoc />
+		public override object Clamp(object value)
+		{
+			if (value is T v && _acceptableValues.Any(x => x.Equals(v)))
```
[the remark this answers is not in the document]
ManlyMarco: Yes, will fix

BepInEx/Configuration/ConfigEntry.cs
```diff
@@ -0,0 +1,247 @@
...
+using System.Linq;
+using BepInEx.Logging;
+
+namespace BepInEx.Configuration
+{
+	/// <summary>
+	/// Container for a single setting of a <see cref="Configuration.ConfigFile"/>. 
+	/// Each config entry is linked to one config file.
+	/// </summary>
+	public sealed class ConfigEntry
```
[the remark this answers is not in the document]
ManlyMarco: Yes, I made it to be easy to implement in config manager, but now that I look at it a better solution to have a base class with `object Value` and a generic class with `T Value`. Will change.

BepInEx/Contract/PluginInfo.cs
ManlyMarco: Just for consistency, everything else in the folder is in BepInEx namespace. It shouldn't break anything since the class instances were not exposed to plugins before.

BepInExTests/Configuration/ConfigFileTests.cs
```diff
@@ -0,0 +1,322 @@
...
+		}
+
+		[TestMethod]
+		public void ValueListTest()
+		{
+			var c = MakeConfig();
+			var w = c.GetSetting("Cat", "Key", "kek", new ConfigDescription("Test", new AcceptableValueList<string>("lel", "kek", "wew", "why")));
+
+			Assert.AreEqual("kek", w.Value);
+			w.Value = "wew";
```
bbepis: Wew

BepInEx/Contract/Attributes.cs
```diff
@@ -296,6 +295,7 @@ public static IEnumerable<BepInDependency> GetDependencies(Type plugin)
 	/// <summary>
 	/// An exception which is thrown when a plugin's dependencies cannot be found.
 	/// </summary>
+	[Obsolete]
```
bbepis: I'm fine with removing the class entirely. IIRC it was only ever used internally in chainloader and I can't see how any plugins would've used this, since they don't have access to chainloader internals

Could argue that some plugins _may_ have used it but I'm willing to bet otherwise

BepInEx/Configuration/TypeConverter.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace BepInEx.Configuration
+{
+	/// <summary>
+	/// A serializer/deserializer combo for some type(s). Used by the config system.
+	/// </summary>
+	public class TypeConverter
```
bbepis: I've talked with horse about using a type that's already in .NET (System.ComponentModel.TypeConverter), however I'm not sure if it's possible to be used like this
https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.typeconverter?view=netframework-4.8

Interested to hear your thoughts on the situation

BepInEx/Configuration/AcceptableValueRange.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace BepInEx.Configuration
+{
+	/// <summary>
+	/// Specify the range of acceptable values for a setting.
+	/// </summary>
+	public sealed class AcceptableValueRange<T> : AcceptableValueBase where T : IComparable
+	{
+		/// <param name="minValue">Lowest acceptable value</param>
+		/// <param name="maxValue">Highest acceptable value</param>
+		public AcceptableValueRange(T minValue, T maxValue) : base(typeof(T))
```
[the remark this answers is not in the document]
ManlyMarco: I don't think it's necessary. It wouldn't be used for anything by this class, so it would just be an arbitrary restriction.

BepInEx/Contract/Attributes.cs
ManlyMarco: Done.

BepInExTests/Configuration/ConfigFileTests.cs
ManlyMarco: Kek

BepInEx/Configuration/ConfigDefinition.cs
```diff
@@ -42,10 +69,22 @@ public override int GetHashCode()
...
+		/// <summary>
+		/// Check if the definitions are the same.
+		/// </summary>
 		public static bool operator !=(ConfigDefinition left, ConfigDefinition right)
 			=> !Equals(left, right);
+
+		/// <inheritdoc />
+		public override string ToString()
+		{
+			return Section + " / " + Key;
```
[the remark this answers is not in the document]
ManlyMarco: Done.

BepInEx.sln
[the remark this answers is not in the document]
ManlyMarco: Will do it as the last commit when the PR is accepted, VS will just keep changing it back otherwise.

BepInEx/Configuration/TypeConverter.cs
ManlyMarco: Dropped a word on discord.

ManlyMarco:
**Overview of the changes**
The changes to the config system aim to make it easier to use by modders and users. Modders get a more concrete API with more features, while users get more information in config files and can use the ConfigurationManager plugin to access and change settings in-game (it works in almost any game under almost any version of Unity. It will get a proper release after these changes are finalized because it heavily depends on them). Plugins made for RC1 will still work (old methods have been marked as obsolete, not removed) and can be updated to the new API whenever convenient.

configman preview

**Detailed changes**
- Config.Wrap has been obsoleted. It has been split into AddSetting (in most cases the same as Wrap, except it can be called only once for a single setting - if you stored the SettingWrapper to a variable it doesn't affect you) and GetSetting (gets a setting if it has been added by AddSetting, best to use when accessing other plugin's settings). This is to make creating settings easier to understand and more predictable - before you could call Wrap multiple times and the description would be ignored after the 1st call. Each call to Wrap also returned a new copy of the ConfigWrapper - now there is only one instance of a setting, and any metadata of the setting cannot be changed after the AddSetting call.
- Specifying the defaultValue is now mandatory in AddSetting to avoid forgetting to set it (it also makes it not necessary to specify <T>).
- Added support for AcceptableValues to config system - in AddSetting you can now specify what values the setting can have, for example (0 to 50) - range, (1, 5, 10 or 15) - list. Invalid values are automatically clamped to nearest valid value, so checking if the setting value is in a valid range is no longer necessary in plugin code. Information about these ranges is added to config files and special UI is used in ConfigurationManager plugin for them.
- Added KeyboardShortcut class for easy-to-use keybinds - arbitrary number of modifier keys are supported, hotkeys with same main key but different modifier keys don't conflict with each other.
- Config system now supports storing enums, newlines in strings, Vector2, Vector3, Vector4, Quaternion, Color and KeyboardShortcut.
- Can now add custom type support to config system by using TomlTypeConverter.AddConverter.
- Config files now have much richer comments with automatically generated text for setting type, default value and value range if any.
- Can now access all of a plugin's settings with PluginInstance.Config.GetConfigEntries(). Mostly useful for setting managers (namely ConfigurationManager).
- Added ThreadingHelper - helps to multithread your code - can run actions on main unity thread or on a background thread, ThreadingHelper.SynchronizingObject can be used with .NET APIs (e.g. timers, filesystemwatcher) to make them automagically synchronize their events (e.g. Elapsed, FileChanged) to the main unity thread.
- Fixed null sections acting weird in config files.
- Removed obsolete MissingDependencyException.

PR #90, opened 2019-08-25
 Added MinimumVersion support to BepInDependency and added BepInIncompatibility

ManlyMarco:
Plugins can require a minimum version of another plugin. 
Exposed DependencyErrors is for displaying ton screen in MessageCenter so average user can see something is wrong.

### DIFF
--- a/BepInEx/Utility.cs
+++ b/BepInEx/Utility.cs
@@ -237,5 +237,10 @@ namespace BepInEx
 				return false;
 			}
 		}
+
+		internal static bool IsZero(this Version v)
+		{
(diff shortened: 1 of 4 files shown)


BepInEx/Utility.cs
```diff
@@ -237,5 +237,10 @@ public static bool TryOpenFileStream(string path, FileMode mode, out FileStream
 				return false;
 			}
 		}
+
+		internal static bool IsZero(this Version v)
```
bbepis: Considering that this is used in one place and I think the use case for this is really limited, it might be a good idea to just have this code explicitly written in the only call, and the utility method removed

PR #103, opened 2019-11-30
Added warning if a plugin was built against a newer version of bepinex

ManlyMarco:


BepInEx/Bootstrap/Chainloader.cs
```diff
@@ -195,6 +198,17 @@ private static bool HasBepinPlugins(AssemblyDefinition ass)
 			return true;
 		}
 
+		private static bool PluginTargetsNewerBepin(PluginInfo pluginInfo)
+		{
+			var pluginTarget = pluginInfo.TargettedBepInExVersion;
+			// X.X.X.x - compare normally. x.x.x.X - nightly build number, ignore
+			if (pluginTarget.Major < CurrentAssemblyVersion.Major) return false;
```
bbepis: Make this an inequality check instead of less-than

PR #123, opened 2020-07-10
Update run_bepinex.sh

notfood:
- Allows paths with spaces
- Will warn and exit if there is no executable, not just the argument, no need to remove any lines after configuring it
- Removed unnecesary ;

### DIFF
--- a/doorstop/run_bepinex.sh
+++ b/doorstop/run_bepinex.sh
@@ -10,7 +10,7 @@
 
 # EDIT THIS: The name of the executable to run
 # LINUX: This is the name of the Unity game executable 
-# MACOS: This is the name of the game app WITHOUT the .app
+# MACOS: This is the name of the game app folder, include the .app suffix
 executable_name=""
 
 # The rest is automatically handled by BepInEx
@@ -41,6 +41,7 @@ case $os_type in
         lib_postfix="so"
         ;;
     Darwin*)
+        executable_name=`basename "${executable_name}" .app`
         executable_path="${PWD}/${executable_name}.app/Contents/MacOS/${executable_name}"
         lib_postfix="dylib"
         ;;
@@ -86,6 +87,6 @@ doorstop_libname=libdoorstop_${arch}.${lib_postfix}
 export LD_LIBRARY_PATH="${doorstop_libs}":${LD_LIBRARY_PATH}
 export LD_PRELOAD=$doorstop_libname:$LD_PRELOAD
 export DYLD_LIBRARY_PATH="${doorstop_libs}"
-export DYLD_INSERT_LIBRARIES=$doorstop_libname
+export DYLD_INSERT_LIBRARIES="${doorstop_libs}/$doorstop_libname"
 
 "${executable_path}"
--- a/doorstop/run_bepinex.sh
+++ b/doorstop/run_bepinex.sh
@@ -5,47 +5,50 @@
 #
 # Usage: Configure the script below and simply run this script when you want to run your game modded.
 
-if [ -z "$1" ]; then echo "Please open run.sh in a text editor and configure executable name. Comment or remove this line when you're done." && exit 1; fi
-
 # -------- SETTINGS --------
 # ---- EDIT AS NEEDED ------
 
 # EDIT THIS: The name of the executable to run
 # LINUX: This is the name of the Unity game executable 
 # MACOS: This is the name of the game app WITHOUT the .app
-executable_name="";
-
+executable_name=""
 
 # The rest is automatically handled by BepInEx
 
 # Whether or not to enable Doorstop. Valid values: TRUE or FALSE
-export DOORSTOP_ENABLE=TRUE;
+export DOORSTOP_ENABLE=TRUE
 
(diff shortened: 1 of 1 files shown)


denikson:
Great, thank you! What OSes was this tested on? Primary special cases we have to look for is

* Linux, running `run_bepinex.sh`
* Linux, running from Steam as per [interop guide](https://bepinex.github.io/bepinex_docs/master/articles/advanced/steam_interop.html) 
* macOS, running `run_bepinex.sh`
* macOS, running Steam as per [interop guide](https://bepinex.github.io/bepinex_docs/master/articles/advanced/steam_interop.html)

notfood:
I have only been able to test
- Linux, running run_bepinex.sh
- Linux, running from Steam

doorstop/run_bepinex.sh
```diff
@@ -5,73 +5,87 @@
...
-
+export DOORSTOP_INVOKE_DLL_PATH="${PWD}/BepInEx/core/BepInEx.Preloader.dll"
 
 # ----- DO NOT EDIT FROM THIS LINE FORWARD  ------
 # ----- (unless you know what you're doing) ------
 
-# Backup current LD_PRELOAD because it can break `file` when running from Steam
-LD_PRELOAD_BAK=$LD_PRELOAD;
-export LD_PRELOAD="";
+if [ ! -x "$1" -a ! -x "$executable_name" ]; then
```
denikson, line 27: This check doesn't work on macOS, because apps have `.app` extension, but `executable_name` specifies to not include `.app` (because normally you don't see this extension in Finder).

I think we can require users to specify `.app` on macOS in the `executable_name` so that the user **must** add `.app` at the end of the `executable_name`. That way the behaviour is the same as launching the game with cmdline args

doorstop/run_bepinex.sh
```diff
@@ -5,73 +5,87 @@
...
 # Usage: Configure the script below and simply run this script when you want to run your game modded.
 
-if [ -z "$1" ]; then echo "Please open run.sh in a text editor and configure executable name. Comment or remove this line when you're done." && exit 1; fi
-
 # -------- SETTINGS --------
 # ---- EDIT AS NEEDED ------
 
 # EDIT THIS: The name of the executable to run
 # LINUX: This is the name of the Unity game executable 
 # MACOS: This is the name of the game app WITHOUT the .app
```
denikson, line 13: I think we can change that on macOS you **must** specify `.app` prefix for executable names, see comment below.

doorstop/run_bepinex.sh
```diff
@@ -5,73 +5,87 @@
...
-                lib_postfix="dylib";;
-    *)          echo "Cannot identify OS (got $(uname -s))!";
-                echo "Please create an issue at https://github.com/BepInEx/BepInEx/issues."; 
-                exit 1;;
+    Linux*)
+        executable_path="${PWD}/${executable_name}"
+        lib_postfix="so"
+        ;;
+    Darwin*)
+        executable_path="${PWD}/${executable_name}.app/Contents/MacOS/${executable_name}"
```
denikson, line 44: Naturally, this line should be changed to remove `.app` if `executable_name` were to include `.app` into itself.

doorstop/run_bepinex.sh
```diff
@@ -5,73 +5,87 @@
...
-doorstop_libname=libdoorstop_${arch}.${lib_postfix};
-export LD_LIBRARY_PATH=${doorstop_libs}:${LD_LIBRARY_PATH};
-export LD_PRELOAD=${doorstop_libs}/$doorstop_libname:$LD_PRELOAD_BAK;
-export DYLD_LIBRARY_PATH=${doorstop_libs};
-export DYLD_INSERT_LIBRARIES=${doorstop_libs}/$doorstop_libname;
+doorstop_libname=libdoorstop_${arch}.${lib_postfix}
+export LD_LIBRARY_PATH="${doorstop_libs}":${LD_LIBRARY_PATH}
+export LD_PRELOAD=$doorstop_libname:$LD_PRELOAD
+export DYLD_LIBRARY_PATH="${doorstop_libs}"
+export DYLD_INSERT_LIBRARIES=$doorstop_libname
```
denikson, line 89: On macOS it seems that you need to specify the full path to the library into `DYLD_INSERT_LIBRARIES` as it does not properly respect `DYLD_LIBRARY_PATH`. Unlike linux, macOS handles spaces fine, so you can revert it to
```sh
export DYLD_LIBRARY_PATH=${doorstop_libs}
export DYLD_INSERT_LIBRARIES=${doorstop_libs}/$doorstop_libname
```

doorstop/run_bepinex.sh
notfood: Small detail, executable_name needs to be without the .app suffix for the last part.

notfood:
Added some changes to resolve the raised issues. I can't test properly against macOS.

denikson:
Perfect, I'll have it tested asap just one more time

denikson:
This is now available in bleeding edge builds

Thank you very much for cleaning up the start script!

PR #138, opened 2020-09-13
Limit line.Split in ConfigFile.Reload()

xoxfaby:
## Description

Limiting the line.Split in Reload() to 2 substrings to allow for values containing `=` characters. Original functionality is maintained.

## Motivation and Context

I stumbled across this bug when trying to use `=` characters in my config values.  
I see no reason config values shouldn't be allowed to contain `=`'s, I assume this was an oversight when originally implemented. 

## How Has This Been Tested?

This is such a small, basic change that I have not tested it but if absolutely necessary I can test it.

## Types of changes

## Checklist:

### DIFF
--- a/BepInEx/Configuration/ConfigFile.cs
+++ b/BepInEx/Configuration/ConfigFile.cs
@@ -118,7 +118,7 @@ namespace BepInEx.Configuration
 						continue;
 					}
 
-					string[] split = line.Split('='); //actual config line
+					string[] split = line.Split(new [] { '=' }, 2); //actual config line
 					if (split.Length != 2)
 						continue; //empty/invalid line
 


BepInEx/Configuration/ConfigFile.cs
```diff
@@ -118,7 +118,7 @@ public void Reload()
 						continue;
 					}
 
-					string[] split = line.Split('='); //actual config line
+					string[] split = line.Split('=',2); //actual config line
```
denikson, line 121: .NET Framework 3.5 does not have an overload `String.Split(char, int)`, so this will not compile. If you go this way, you'd have to pick [an overload that's available in FW 3.5](https://docs.microsoft.com/en-us/dotnet/api/system.string.split?view=netframework-3.5) as that's the target BepInEx uses.

denikson (changes requested):
Greetings! 

Thank you for the pull request. I've added an important note regarding the change, as it will not compile now.

However, there is a bigger issue at hand: what if having `=` is expected input for a value? In my opinion, a better option is to pick the first `=` and include the rest to be part of the value (which is doable with a `IndexOf` and a `Substring`). While still not perfect, it would be a good enough workaround until 6.0 where we are intending on using full TOML parser.

xoxfaby:
Oh I see, my bad.

As for the other comment, if String.Split is limited, the last substring will contain the rest of the string, that was the intention with this change. 

`String.Split(['=`], 2)` should compile and work as you described.

denikson:
> As for the other comment, if String.Split is limited, the last substring will contain the rest of the string, that was the intention with this change.

Ah, that is indeed true. In that case just a split should work.

> `String.Split(['='], 2)` should compile and work as you described.

That's still not valid C#; to define an array you would use `new []{ '=' }`.

Would you mind pushing the change so I can merge the PR or are you fine with me making a manual push to master directly?

xoxfaby:
Derp, thank you, my brain is still in python land sometimes. Now it should be all good?

denikson:
Yes, looks good to me!
Thank you for the PR again!

PR #155, opened 2020-11-16
Update cake script to use msbuild instead of AssemblyInfo.cs

js6pak:


build.cake
```diff
@@ -53,26 +53,29 @@ Task("Build")
...
+        };
 
-    buildVersion = FindRegexMatchInFile(bepinExProperties + File("VersionInfo.cs"), "[0-9]+\\.[0-9]+\\.[0-9]+\\.[0-9]+", System.Text.RegularExpressions.RegexOptions.None);
+        buildSettings.Properties["AssemblyVersion"] = new[] { buildVersion + "." + buildId };
 
-    var buildSettings = new MSBuildSettings {
-        Configuration = "Release",
-        Restore = true
-    };
+        buildVersion += "-ci." + buildId;
```
denikson, line 75: We call our CI builds "bleeding edge", so I think it's better to instead add `-be.` pre-release tag instead.

denikson (changes requested):
Looks good to me! Added a single change request. Once it's handled, I'll merge this in.

PR #167, opened 2021-01-09
Add ability to disable generating setting descriptions in ConfigFile

ManlyMarco:


BepInEx.Core/Configuration/ConfigFile.cs
```diff
@@ -93,6 +93,11 @@ public ConfigFile(string configPath, bool saveOnInit, BepInPlugin ownerMetadata)
 		#region Save/Load
 
 		private readonly object _ioLock = new();
+		
+		/// <summary>
+		/// Generate user-readable comments for each of the settings in the saved .cfg file.
+		/// </summary>
+		public bool GenerateSettingDescriptions { get; set; } = true;
```
denikson, line 100: This may change down the road depending on where we will take the config system, but for this whis is fine.

denikson (approved):

PR #171, opened 2021-01-10
Hash unity base libs and move them to BepInEx/unity-libs

js6pak:
## Description
Hash unity base libs so unhollowed assemblies get regenerated when base libs are changed.
Move base libs to `BepInEx/unity-libs` so `BepInEx/unhollowed` can be safely removed again.

## Types of changes

## Checklist:

### DIFF
--- a/BepInEx.IL2CPP/ProxyAssemblyGenerator.cs
+++ b/BepInEx.IL2CPP/ProxyAssemblyGenerator.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.IO;
 using System.Security.Cryptography;
+using System.Text;
 using AssemblyUnhollower;
 using BepInEx.Logging;
 using HarmonyLib;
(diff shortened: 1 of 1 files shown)


BepInEx.IL2CPP/ProxyAssemblyGenerator.cs
```diff
@@ -42,19 +43,34 @@ public static AppDomain CreateDomain(string name, AppDomainSetup setup)
...
+			{
+				foreach (var file in Directory.EnumerateFiles(UnityBaseLibsDirectory, "*.dll", SearchOption.TopDirectoryOnly))
+				{
+					var pathBytes = Encoding.UTF8.GetBytes(Path.GetFileName(file));
+					md5.TransformBlock(pathBytes, 0, pathBytes.Length, pathBytes, 0);
 
-			return Utility.ByteArrayToString(hash);
+					var contentBytes = File.ReadAllBytes(file);
+					md5.TransformBlock(contentBytes, 0, contentBytes.Length, contentBytes, 0);
+				}
```
denikson, line 68: Did you measure what the inpact on load times with this? Or at least how long this takes on average?

BepInEx.IL2CPP/ProxyAssemblyGenerator.cs
```diff
@@ -157,26 +173,13 @@ public void GenerateAssembliesInternal(AppDomainListener listener)
 				UnhollowerBaseLib.LogSupport.TraceHandler += s => listener.DoUnhollowerLog(s, LogLevel.Debug);
 				UnhollowerBaseLib.LogSupport.ErrorHandler += s => listener.DoUnhollowerLog(s, LogLevel.Error);
 
-
-				string unityBaseLibDir = Path.Combine(Preloader.IL2CPPUnhollowedPath, "base");
-
-				if (Directory.Exists(unityBaseLibDir))
-				{
-					listener.DoPreloaderLog("Found base unity libraries", LogLevel.Debug);
```
denikson, line 165: The log message is still useful for technical support as it will tell if the user has base unity libraries.
This should be restored at some place in the code.

js6pak: https://github.com/knah/Il2CppAssemblyUnhollower/blob/f7a2cd63dbc574f78c0e7944b53876365af2a317/AssemblyUnhollower/Program.cs#L222

denikson: That doesn't appear to use Unhollower's `LogSupport` logging system and instead print it all to stdout, which is not captured by disk logger of BepInEx owing to stdout being redirected to the console instead. 

As such, one option here is to make a PR to knah/Il2CppAssemblyUnhollower and use `LogSupport` like in [here for example](https://github.com/knah/Il2CppAssemblyUnhollower/blob/f7a2cd63dbc574f78c0e7944b53876365af2a317/AssemblyUnhollower/Program.cs#L184) or to manually output this on BepInEx side so that this important info ends up into the disk log as well.

Issue #229, opened 2021-04-16
IL2CPP Preloader - fatal error when getting unity version info


BlacRyu:
**Describe the bug**
I'm getting a fatal error in the preloader when launching the game SaGa Frontier Remastered with BepInEx. It seems there is some extra information being passed through when it tries to access the file version info from the .NET API.

**Screenshots and logs**
[LogOutput.log](https://github.com/BepInEx/BepInEx/files/6328008/LogOutput.log)
![image](https://user-images.githubusercontent.com/10762433/115087275-b73cef00-9ec2-11eb-9c3a-8202a416d99e.png)
In this screenshot I used a custom build to print the full file version info. You can see extra data gets added to all the file version info string properties.

**Desktop (please complete the following information):**
 - Game and game version: SaGa Frontier Remastered v1.0 (Steam build [6443908](https://steamdb.info/patchnotes/6443908/))
 - BepInEx version: 6.0.0 x64 Bleeding Edge 361

Kein:
This can be addressed with basic string version slice to 10-11 chars, since for Version Parse you only need  major, minor and build, however, this raises the question what would you do if version is stripped? You can fallback to UnityPlayer module version lookup, but what if it is stripped from there too? 

I think as a final failure it worth to expose config string where user can specify custom version and instead catch exception in version parsing and print the info about config option.

PR #230, opened 2021-04-17
Issue #229 fix

Albeoris:
## Description
Fixed issue #229.
* Now we manually parse the first three components of the version.
* Now we can override the version in the config file: `[Unity] Version = 2019.4.16`
* Moved version initialization to Preloader to avoid confusion in the logs.

### DIFF
--- a/BepInEx.IL2CPP/Preloader.cs
+++ b/BepInEx.IL2CPP/Preloader.cs
@@ -40,12 +40,10 @@ namespace BepInEx.IL2CPP
                 }
 
                 ChainloaderLogHelper.PrintLogInfo(Log);
-                InitializeUnityVersion();
 
                 Log.LogDebug($"Game executable path: {Paths.ExecutablePath}");
                 Log.LogDebug($"Unhollowed assembly directory: {IL2CPPUnhollowedPath}");
                 Log.LogDebug($"BepInEx root path: {Paths.BepInExRootPath}");
-
 
                 UnhollowerLog = Logger.CreateLogSource("Unhollower");
                 LogSupport.InfoHandler += UnhollowerLog.LogInfo;
@@ -56,6 +54,9 @@ namespace BepInEx.IL2CPP
 
                 if (ProxyAssemblyGenerator.CheckIfGenerationRequired())
                     ProxyAssemblyGenerator.GenerateAssemblies();
+                
+                
+                InitializeUnityVersion();
 
 
                 using (var assemblyPatcher = new AssemblyPatcher())
@@ -91,13 +92,8 @@ namespace BepInEx.IL2CPP
         {
             try
             {
-                ConfigEntry<String> config = ConfigFile.CoreConfig.Bind(
-                    "Unity", "Version",
-                    String.Empty,
-                    "Allows you to manually set the Unity version .");
-
-                String version = config.Value;
-                if (String.IsNullOrWhiteSpace(version))
+                var version = ConfigUnityVersion.Value;
+                if (string.IsNullOrWhiteSpace(version))
                 {
                     version = //Version.Parse(Application.unityVersion);
                         Process.GetCurrentProcess().MainModule.FileVersionInfo.FileVersion;
@@ -109,17 +105,20 @@ namespace BepInEx.IL2CPP
                     PreloaderLogger.Log.LogDebug($"Unity version obtained from the config: [{version}]");
                 }
 
-                String[] parts = version.Split('.');
-                Int32 major = 0;
-                Int32 minor = 0;
-                Int32 build = 0;
+                var parts = version.Split('.');
+                var major = 0;
+                var minor = 0;
+                var build = 0;
 
                 // Issue #229 - Don't use Version.Parse("2019.4.16.14703470L&ProductVersion")
-                major = Int32.Parse(parts[0], NumberStyles.Integer, CultureInfo.InvariantCulture);
-                if (parts.Length > 1)
-                    minor = Int32.Parse(parts[1], NumberStyles.Integer, CultureInfo.InvariantCulture);
-                if (parts.Length > 2)
-                    build = Int32.Parse(parts[2], NumberStyles.Integer, CultureInfo.InvariantCulture);
+                bool success = int.TryParse(parts[0], NumberStyles.Integer, CultureInfo.InvariantCulture, out major);
+                if (success && parts.Length > 1)
+                    success = int.TryParse(parts[1], NumberStyles.Integer, CultureInfo.InvariantCulture, out minor);
+                if (success && parts.Length > 2)
+                    success = int.TryParse(parts[2], NumberStyles.Integer, CultureInfo.InvariantCulture, out build);
+
+                if (!success)
+                    throw new InvalidDataException($"Failed to parse Unity version: {version}");
 
                 UnityVersionHandler.Initialize(major, minor, build);
                 Log.LogInfo($"Running under Unity v{major}.{minor}.{build}");
@@ -129,5 +128,14 @@ namespace BepInEx.IL2CPP
                 throw new InvalidDataException("Failed to parse Unity version.", ex);
             }
         }
+        
+        #region Config
+
+        private static readonly ConfigEntry<string> ConfigUnityVersion = ConfigFile.CoreConfig.Bind(
+            "IL2CPP", "UnityVersion",
+            string.Empty,
+            "Unity version to report to Il2CppUnhollower. If empty, version is automatically determined from the game process.");
+        
+        #endregion
     }
 }
--- a/BepInEx.IL2CPP/Preloader.cs
+++ b/BepInEx.IL2CPP/Preloader.cs
@@ -91,11 +91,11 @@ namespace BepInEx.IL2CPP
         private static void InitializeUnityVersion()
         {
             if (TryInitializeUnityVersion(ConfigUnityVersion.Value))
-                Log.LogDebug($"Unity version obtained from the config.");
+                Log.LogWarning($"Unity version obtained from the config.");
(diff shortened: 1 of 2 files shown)


BepInEx.IL2CPP/Preloader.cs
```diff
@@ -83,5 +86,48 @@ public static void Run()
...
         }
+
+        private static void InitializeUnityVersion()
+        {
+            try
+            {
+                ConfigEntry<String> config = ConfigFile.CoreConfig.Bind(
+                    "Unity", "Version",
+                    String.Empty,
+                    "Allows you to manually set the Unity version .");
```
ghorsington, line 97: I suggest moving this to a static readonly field as we do with other config values.

Moreover, I think the config option should be described better as its use is mainly for Il2Cpp. I suggest the following values:

* Section: `IL2CPP`
* Name: `UnityVersion`
* Description: `Unity version to report to Il2CppUnhollower. If empty, version is automatically determined from the game process.`

BepInEx.IL2CPP/Preloader.cs
```diff
@@ -83,5 +86,48 @@ public static void Run()
...
             }
         }
+
+        private static void InitializeUnityVersion()
+        {
+            try
+            {
+                ConfigEntry<String> config = ConfigFile.CoreConfig.Bind(
+                    "Unity", "Version",
+                    String.Empty,
```
ghorsington, line 96: Please use first-class type `string` instead of `String` class.

BepInEx.IL2CPP/Preloader.cs
```diff
@@ -83,5 +86,48 @@ public static void Run()
...
+                    "Allows you to manually set the Unity version .");
+
+                String version = config.Value;
```
ghorsington, line 99: Please use first-class type `string` instead of `String` class.

BepInEx.IL2CPP/Preloader.cs
```diff
@@ -83,5 +86,48 @@ public static void Run()
...
+                if (String.IsNullOrWhiteSpace(version))
```
ghorsington, line 100: Please use first-class type `string` instead of `String` class.

BepInEx.IL2CPP/Preloader.cs
```diff
@@ -83,5 +86,48 @@ public static void Run()
...
+                {
+                    version = //Version.Parse(Application.unityVersion);
+                        Process.GetCurrentProcess().MainModule.FileVersionInfo.FileVersion;
+
+                    PreloaderLogger.Log.LogDebug($"Unity version obtained from main application module: [{version}]");
+                }
+                else
+                {
+                    PreloaderLogger.Log.LogDebug($"Unity version obtained from the config: [{version}]");
```
ghorsington, line 109: Instead of Debug logging here I'd make a Warning level log and report that the Unity version is read from config. This will help to detect bad configuration.

BepInEx.IL2CPP/Preloader.cs
```diff
@@ -83,5 +86,48 @@ public static void Run()
...
+                }
+
+                String[] parts = version.Split('.');
```
ghorsington, line 112: Please use first-class type `string` instead of `String` class.

BepInEx.IL2CPP/Preloader.cs
```diff
@@ -83,5 +86,48 @@ public static void Run()
...
+                Int32 major = 0;
+                Int32 minor = 0;
+                Int32 build = 0;
+
+                // Issue #229 - Don't use Version.Parse("2019.4.16.14703470L&ProductVersion")
+                major = Int32.Parse(parts[0], NumberStyles.Integer, CultureInfo.InvariantCulture);
+                if (parts.Length > 1)
+                    minor = Int32.Parse(parts[1], NumberStyles.Integer, CultureInfo.InvariantCulture);
+                if (parts.Length > 2)
+                    build = Int32.Parse(parts[2], NumberStyles.Integer, CultureInfo.InvariantCulture);
```
ghorsington, line 122: What if the user mistakingly gives a non-integer? That will cause a fail that will not initialize Unhollower properly. I'd suggest either using `int.TryParse` or better yet, define a RegEx here instead.
A RegEx to me looks like a better solution here overall as it will both check the validity and will allow extracting the version number from more complex strings.

BepInEx.IL2CPP/Preloader.cs
```diff
@@ -83,5 +86,48 @@ public static void Run()
...
+                    PreloaderLogger.Log.LogDebug($"Unity version obtained from the config: [{version}]");
+                }
+
+                String[] parts = version.Split('.');
+                Int32 major = 0;
+                Int32 minor = 0;
+                Int32 build = 0;
+
+                // Issue #229 - Don't use Version.Parse("2019.4.16.14703470L&ProductVersion")
+                major = Int32.Parse(parts[0], NumberStyles.Integer, CultureInfo.InvariantCulture);
```
ghorsington, line 118: Use first-class `int` instead of class-based `Int32`

BepInEx.IL2CPP/Preloader.cs
```diff
@@ -83,5 +86,48 @@ public static void Run()
...
+                    minor = Int32.Parse(parts[1], NumberStyles.Integer, CultureInfo.InvariantCulture);
+                if (parts.Length > 2)
+                    build = Int32.Parse(parts[2], NumberStyles.Integer, CultureInfo.InvariantCulture);
+
+                UnityVersionHandler.Initialize(major, minor, build);
+                Log.LogInfo($"Running under Unity v{major}.{minor}.{build}");
+            }
+            catch (Exception ex)
+            {
+                throw new InvalidDataException("Failed to parse Unity version.", ex);
```
ghorsington, line 129: I don't think we should rethrow here. You still should be able to at least launch the game without massive exceptions. A better option might be to add a clear Error level log.

BepInEx.IL2CPP/Preloader.cs
```diff
@@ -36,8 +40,7 @@ public static void Run()
                 }
 
                 ChainloaderLogHelper.PrintLogInfo(Log);
-
-                Log.LogInfo($"Running under Unity v{FileVersionInfo.GetVersionInfo(Paths.ExecutablePath).FileVersion}");
+                InitializeUnityVersion();
```
ghorsington, line 43: If this is run here, Il2CppUnhollower runtime library will be initialized before unhollower DLLs are generated. This may cause issues as the unhollower was designed to work in in two steps:

1. Generate unhollowed assemblies
2. Initialize runtime library based on the generated unhollowed assemblies.

As such, this should be run after `ProxyAssemblyGenerator.GenerateAssemblies()` call.

ghorsington (changes requested):
Thanks for the PR! Here are some changes that I think should be done. While some of the technical choices are up for debate (feel free to comment), ones related to style must be addressed.

BepInEx.IL2CPP/Preloader.cs
```diff
@@ -83,5 +86,48 @@ public static void Run()
                 throw;
             }
         }
+
+        private static void InitializeUnityVersion()
+        {
+            try
+            {
+                ConfigEntry<String> config = ConfigFile.CoreConfig.Bind(
```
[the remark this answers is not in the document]
js6pak: Shouldn't `.editorconfig` warn you to use `var`s anyway? Also please use `var`s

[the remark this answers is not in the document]
Albeoris: Should I use `var` if the left side of the expression isn't obvious from the right side?

BepInEx.IL2CPP/Preloader.cs
Albeoris: I think if the user was typing wrong, we should crash with an error, shouldn't we?

BepInEx.IL2CPP/Preloader.cs
Albeoris: Why? If we initialize the wrong version, it can be difficult to diagnose errors due to incorrect initialization. Now the game crashes in this case, it seems to me that this is the correct behavior (otherwise we would not even know about the linked error).

BepInEx.IL2CPP/Preloader.cs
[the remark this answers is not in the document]
ghorsington: [Yes, as per the coding style we use.](https://github.com/BepInEx/BepInEx/blob/master/.editorconfig#L73-L75)

In general, I still suggest moving the config itself into a static readonly prop like we do with our other config entries.

BepInEx.IL2CPP/Preloader.cs
ghorsington: If the user types wrong, I'd rather warn and initialize Unhollower with its default Unity version-specific handlers.
I'd err on the side of letting Unhollower decide the version handler to use if no Unity version is given, since Unhollower itself has that ability.

If the value is wrong, I'd consider resetting it like we do for other config options.

BepInEx.IL2CPP/Preloader.cs
ghorsington: Fair point. I'm planning on rewriting the version checker either way, so this can be like this for now.

BepInEx.IL2CPP/Preloader.cs
Albeoris: Done.
But I didn't use RegEx - the optional groups (2019 / 2019.4 / 2019.4.16 / 2019.4.16.1) only add complexity to the code, and don't make it easier to read.

Albeoris:
Fixed.

ghorsington (approved):
LGTM

ghorsington:
Thanks for the PR!

PR #243, opened 2021-05-01
Custom Save and Reload implementations for ConfigFile

Algorithman:
## Description
Made ConfigFile class more versatile with virtual Save and Reload methods.
The type to use for the ConfigFile implementation can be added as BepInConfigType attribute to the BaseUnityPlugin class.

## Motivation and Context
More and more games go online, so we had to work around issues involving server side only configuration values, but still keep them all inside one ConfigFile instance.
This not also addresses this, it also addresses others who might want to use their own file format.
The custom implementation can be done overriding SaveImplementation() and ReloadImplementation() methods.

## How Has This Been Tested?
Checked it inside our Valheim plugin, our ConfigFile implementation is instanced correctly and loaded/saved also correctly.

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.

### DIFF
--- a/BepInEx.Core/Configuration/ConfigFile.cs
+++ b/BepInEx.Core/Configuration/ConfigFile.cs
@@ -264,42 +264,50 @@ namespace BepInEx.Configuration
         {
             lock (_ioLock)
             {
-                OrphanedEntries.Clear();
-
-                var currentSection = string.Empty;
-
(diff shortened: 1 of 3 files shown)


ghorsington:
The extendable configuration files are now being implemented as part of #267 and its general configuration overhaul

PR #284, opened 2021-07-26
Make stacktraces Error instead of Debug

js6pak:
Currently stacktraces are logged with Debug log level which is not enabled by default making logs sent by users useless, this fixes it


ManlyMarco (changes requested):
I think most of these should be warnings at most since they aren't critical. Only line 360 should be an error. If the traces are going to be changed from debug level then they should be combined with the rest of the error message.

js6pak:
If plugin fails to load, imo it is pretty critical

ManlyMarco (approved):

PR #351, opened 2022-01-08
CI: Emphasize architecture of game instead of OS

miniduikboot:
## Description

The current wording on the download page states that the architecture of the OS (32 or
64-bit) is leading for which version of BepInEx to download. This
however should be the architecture for which the game is built, as you
can run 32-bit Windows games on 64-bit Windows systems and need the
32-bit version of BepInEx in this case.

## Motivation and Context

I helped an user today who was very confused about why his installation of BepInEx for a 32-bit game was not working. He used 64-bit Windows, so he downloaded the 64-bit version as instructed by the download page. It took a few attempts to convince him to download the 32-bit version. This hopefully reduces the amount of new souls that download the wrong version.

## How Has This Been Tested?

This has not been tested, as I don't have access to the CI environment. I hope that editing the strings in this build script does not make other parts fall over.

## Types of changes

## Checklist:

Maybe there is a way to migrate the current builds over to the new wording?

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -285,23 +285,23 @@ Task("Pack")
                 ["artifacts"] = new Dictionary<string, object>[] {
                     new() {
                         ["file"] = $"BepInEx_UnityMono_x64{commitPrefix}{buildVersion}.zip",
-                        ["description"] = "BepInEx Unity Mono for Windows x64 machines"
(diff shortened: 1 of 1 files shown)


ghorsington:
The CI passed so looks fine to me. Thanks for the PR!

PR #475, opened 2022-08-24
Add chainloader events

js6pak:
Adds events to chainloader to allow for library plugins to make apis like [PluginSingleton](https://github.com/NuclearPowered/Reactor/blob/master/Reactor/PluginSingleton.cs) or [RegisterCustomRpcAttribute](https://github.com/NuclearPowered/Reactor/blob/master/Reactor/RegisterCustomRpcAttribute.cs)


ghorsington (approved):
Going to accept this now, but this is likely to change in future. Importantly:

* Instead of just `Action` it's probably better to pass a proper `EventHandler<T>` in case parameters change in future
* `PluginLoad` is in Il2Cpp but not in other runtimes that use the `BasePlugin` approach.

PR #477, opened 2022-08-28
Warn about outdated wine

js6pak:


ghorsington (approved):
Looks fine to me for now.

In future this might be better adjusted to use [`PlatformUtils.WineVersion`](https://github.com/BepInEx/BepInEx/blob/master/BepInEx.Preloader.Core/PlatformUtils.cs#L13) instead since it's already provided and logged.

PR #496, opened 2022-10-09
fix: handle parameter spread for lts version

Duelsik:
## Description
Allows the modded game to be ran with parameters from launcher level.
basically it allows to do something like this:
`/bin/sh "/pat/to/game/directory/run_bepinex.sh" "" "-screen-width 1000"` after setting cwd to game directory.
... this `""` works like skipping argument for game executable AND doesn't appear in $@ from game point of view

## Motivation and Context
I am rewriting CastleStory launcher and figured out that this is most likely a mistake in run_bepinex, I don't want to update to 6.x yet - it is fixed there tho (at least looks like so).

## How Has This Been Tested?
I've just run the game from command line and by spawning process from launcher.

## Types of changes

- [x] (I guess?) New feature (non-breaking change which adds functionality)

## Checklist:
- [?] My change requires a change to the documentation.
- [?] I have updated the documentation accordingly.

(I can update docs if it is needed ^^)

### DIFF
--- a/doorstop/run_bepinex.sh
+++ b/doorstop/run_bepinex.sh
@@ -142,4 +142,4 @@ export LD_PRELOAD=$doorstop_libname:$LD_PRELOAD
 export DYLD_LIBRARY_PATH="${doorstop_libs}"
 export DYLD_INSERT_LIBRARIES="${doorstop_libs}/$doorstop_libname"
 
-"${executable_path}" $@
+"${executable_path}" "$@"
--- a/doorstop/run_bepinex.sh
+++ b/doorstop/run_bepinex.sh
@@ -142,4 +142,4 @@ export LD_PRELOAD=$doorstop_libname:$LD_PRELOAD
 export DYLD_LIBRARY_PATH="${doorstop_libs}"
 export DYLD_INSERT_LIBRARIES="${doorstop_libs}/$doorstop_libname"
 
-"${executable_path}"
+"${executable_path}" $@


bb010g:
Please double quote `$@`. If you need to `shift` to make that work, that's fine, but don't expand unquoted variables.

Duelsik:
Thank you, I didn't thought about that. I will push the fix later today.

Duelsik:
This looks interesting as additional safety measure on top of quoting
(sfw) https://dev.to/greymd/why-1-is-used-in-shell-script-364h

Duelsik:
For readability sake I've sticked to `"$@"` instead of what this dev.to article suggests. (+in the article they say that it solves issues that are long gone)
On my side - I had to change the call to be more like
`spawn(binsh, [ pathToScript, "", "-flag", flagvalue ]);`
Instead of passing everything in one string :)
Tested with and without params, in both - terminal and launcher.

cc @bb010g

bb010g:
Thanks! If that's all that's needed to make your usage work, that's great. <https://mywiki.wooledge.org/BashFAQ> and <https://mywiki.wooledge.org/BashPitfalls> are wonderful resources when writing shell, by the way.

bb010g (approved):

Duelsik:
:eyes: Oh, it will help me, thanks for the recommendation :tada: 
For me - ready to merge, but don't have perms to click "the button" ofc ^^

Duelsik:
Hi @bb010g - do you know who I would have to ping to get that merged?

ghorsington (approved):

PR #510, opened 2022-11-11
Add TomlTypeConverter for UnityEngine.Rect

ManlyMarco:
For some reason this was never added to mainline, the converter has been in use for a long time by other plugins adding it.

### DIFF
--- a/BepInEx/Configuration/TomlTypeConverter.cs
+++ b/BepInEx/Configuration/TomlTypeConverter.cs
@@ -341,6 +341,50 @@ namespace BepInEx.Configuration
 			TomlTypeConverter.AddConverter(typeof(Vector3), jsonConverter);
 			TomlTypeConverter.AddConverter(typeof(Vector4), jsonConverter);
 			TomlTypeConverter.AddConverter(typeof(Quaternion), jsonConverter);
+			
+			TomlTypeConverter.AddConverter(typeof(Rect), new TypeConverter
+			{
+				ConvertToObject = delegate (string s, Type type)
+				{
+					Rect result = default(Rect);
+					if (s != null)
+					{
+						string cleaned = s.Trim(new char[] { '{', '}' }).Replace(" ", "");
+						foreach (string part in cleaned.Split(new char[] { ',' }))
+						{
+							string[] parts = part.Split(new char[] { ':' });
+							float value;
+							if (parts.Length == 2 && float.TryParse(parts[1], out value))
+							{
+								string id = parts[0].Trim(new char[] { '"' });
+								if (id == "x")
+								{
+									result.x = value;
+								}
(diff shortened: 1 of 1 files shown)


BepInEx/Configuration/TomlTypeConverter.cs
```diff
@@ -341,6 +341,50 @@ public static void AddUnityEngineConverters()
...
+								}
+								else if (id == "height" || id == "w")
+								{
+									result.height = value;
+								}
+							}
+						}
+					}
+
+					return result;
```
ghorsington, line 380: Maybe move this into a separate function to keep `AddUnityEngineConverters` readable?

BepInEx/Configuration/TomlTypeConverter.cs
```diff
@@ -341,6 +341,50 @@ public static void AddUnityEngineConverters()
...
+								}
+								else if (id == "width" || id == "z")
+								{
+									result.width = value;
+								}
+								else if (id == "height" || id == "w")
+								{
+									result.height = value;
+								}
+							}
```
ghorsington, line 376: I suppose this is just JSON? Maybe switch to use [`JsonUtility.FromJson`](https://docs.unity3d.com/ScriptReference/JsonUtility.FromJson.html)?

ManlyMarco: If I remember correctly using it on Rect doesn't work in all unity versions.

BepInEx/Configuration/TomlTypeConverter.cs
ManlyMarco: Did it.

BepInEx/Configuration/TomlTypeConverter.cs
ghorsington: Hmm, fair enough

BepInEx/Configuration/TomlTypeConverter.cs
```diff
@@ -341,6 +341,46 @@ public static void AddUnityEngineConverters()
...
+			TomlTypeConverter.AddConverter(typeof(Rect), new TypeConverter
+			{
+				ConvertToObject = StringToRect,
+				ConvertToString = RectToString
+			});
+		}
+
+		private static object StringToRect(string s, Type type)
+		{
+			Rect result = default(Rect);
```
ghorsington, line 354:
```suggestion
			// JsonUtility doesn't work with Rect on all Unity versions, so parse it manually

			Rect result = default(Rect);
```

ghorsington: How about a note like this?

ghorsington (approved):

PR #546, opened 2023-01-24
Add an useragent when downloading unity libraries

miniduikboot:
## Description

This adds an user agent to the download request, which is configured to be `BepInEx/` + the version of BepInEx in use.

Reporting an OS or pretending to be [an actual browser](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) didn't feel right

## Motivation and Context

Currently users in some countries are blocked from downloading unity libs, although they are able to download the libraries using their browser. Because of this I'm assuming that the combination of country and the lack of an user agent is causing the download to fail. Add an user agent identifying us as BepInEx and hope that makes the request less sus.

This hopefully fixes what I've been calling the Kazakhstan problem, as reported by me in April last year: https://discord.com/channels/623153565053222947/624272422295568435/968527472377344070 . This issue doesn't appear frequently, but in support channels we only see the tip of the iceberg. So far I've seen users with issues in Kazakhstan (3), Pakistan (1) and Russia (1). 

## How Has This Been Tested?

- I tested it against the original server => still works
- I tested it against a local cache to verify it actually sets an user agent => see screenshot for a before/after 
- I'm reaching out to a user in one of the affected countries to see if this lets them download from the original server => WIP

Http server log of the results of step 2:
2023-01-24T20:56:22,000477055+01:00

## Types of changes

## Checklist:

### DIFF
--- a/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
+++ b/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
@@ -246,6 +246,10 @@ internal static partial class Il2CppInteropManager
             Directory.EnumerateFiles(UnityBaseLibsDirectory, "*.dll").Do(File.Delete);
 
             using var httpClient = new HttpClient();
+            var bepinVersion = Paths.BepInExVersion;
+            var version = new SemanticVersioning.Version(bepinVersion.Major, bepinVersion.Minor, bepinVersion.Patch,
+                                                         bepinVersion.PreRelease);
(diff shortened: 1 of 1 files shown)


ManlyMarco:
Did this end up fixing the download issue for the affected users?

ManlyMarco:
I had this tested on multiple occasions, including just now with a fake firefox useragent, and it didn't seem to help with the download issue for affected users.

ManlyMarco:
Closing since this did not work to fix the issue and there doesn't seem to be any reason to merge.

PR #556, opened 2023-02-15
Expose a DisplayBepInExVersion

js6pak:
## Description
Moves the stripping of version build info from `ChainloaderLogHelper` to a public property

As @ghorsington [noted](https://github.com/BepInEx/BepInEx/blame/6aabdb5ff9500fca541dbd66df7fbd7fe1dacd78/BepInEx.Core/Paths.cs#L14) it probably shouldn't be in `Paths` though, I don't see any existing type good for this, so where should this be?

## Motivation and Context
#546 and [mods that might want to display the bepinex version](https://github.com/NuclearPowered/Reactor/blob/4deb5f88b904a9bd964938ee2ed0a4925f52d830/Reactor/Patches/ReactorVersionShower.cs#L76)


ManlyMarco (commented):
If the build version is only meant to be used in one place then why not just strip the build number there? Also personally I'd rather keep the build number visible in the console title but that's just me.

Edit: Ah, if other plugins want to use this version then why not add an extension method for stripping the build number? Having a second barely different version property just seems like an overkill to me.

BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -111,7 +111,7 @@ protected static bool PluginTargetsWrongBepin(PluginInfo pluginInfo)
 
     #region Contract
 
-    protected virtual string ConsoleTitle => $"BepInEx {Paths.BepInExVersion} - {Paths.ProcessName}";
+    protected virtual string ConsoleTitle => $"BepInEx {Paths.DisplayBepInExVersion} - {Paths.ProcessName}";
```
ManlyMarco: Revert the behaviour change to only leave the new property in this PR, so it can finally be merged.
```suggestion
    protected virtual string ConsoleTitle => $"BepInEx {Paths.BepInExVersion} - {Paths.ProcessName}";
```

BepInEx.Core/BepInEx.Core.csproj
```diff
@@ -5,6 +5,7 @@
         <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
         <!-- Force nuget assembly output on netstandard2.0 -->
         <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+        <RootNamespace>BepInEx</RootNamespace>
```
ManlyMarco:
```suggestion
```

ManlyMarco (approved):

Issue #559, opened 2023-02-24
SteamLaunch doesn't match parameters on Linux


Baerbeisser:
### What happened?

This part

```
if [ "$2" = "SteamLaunch" ]; then
    "$1" "$2" "$3" "$4" "$0" "$5"
    exit
fi
```

tries to run the script with steam-launch-wrapper, instead of the executable.
At least that's what the output suggest.
Did the format change? It looks like this:

```
.../.local/share/Steam/ubuntu12_32/reaper
SteamLaunch
AppId=892970
--
.../.local/share/Steam/ubuntu12_32/steam-launch-wrapper
--
.../.local/share/Steam/steamapps/common/Valheim/valheim.x86_64
```

Consequently, nothing runs.
What is the purpose of this anyway, should it replace run_bepinex.sh ($0) with steam-launch-wrapper?

But even running the script alone, no config, nothing happens. I get the message `Found UnityPlayer, hooking into it instead`, might be related to this?

### Steps to reproduce

Preparing Bepinex like instructed [here](https://docs.bepinex.dev/articles/advanced/steam_interop.html).
Add somewhere in run_bepinex.sh `echo "$@" > command.txt` and compare it with the script.

### BepInEx Distribution

Stable from GitHub

### Environment

```markdown
- OS: Artix (Arch derivative), up to date
- BepInEx: 5.4.21 LTS, unix variant
- Game: Valheim
```

zeroepoch:
I ran into this same problem. I don't know if it's an update to Steam recently that is causing this issue or something else. I believe the extra `--` arguments are messing up the argument counts for the relaunch. Using the following change I was able to get the game working. Basically added 2 extra arguments and shifted the last argument by 2.

```
if [ "$2" = "SteamLaunch" ]; then
    "$1" "$2" "$3" "$4" "$5" "$6" "$0" "$7"
    exit
fi
```

Baerbeisser:
That's what i tried too, among other combinations. Tried it again with your variant, this is the console output i got:

```
Game process added : AppID 892970 "'~/.local/share/Steam/steamapps/common/Valheim/run_bepinex.sh' ~/.local/share/Steam/ubuntu12_32/reaper SteamLaunch AppId=892970 -- ~/.local/share/Steam/ubuntu12_32/steam-launch-wrapper -- '~/.local/share/Steam/steamapps/common/Valheim/valheim.x86_64'", ProcID 8134, IP 0.0.0.0:0
[...]

PR #391, opened 2022-03-24
Restructure project, add Doorstop 4 support, use CoreCLR for Il2Cpp

ghorsington:
Migrated from #385 so that the branch now resides in this repo.

Add Doorstop 4 support, replacing older UnityDoorstop 3. This includes also upgrading Il2Cpp to use self-hosted CoreCLR runtime as that's what Doorstop 4 uses.

## Motivation and Context
Debuggers are handy to have.

## How Has This Been Tested?
I've tested it with two games so far, StuntMANIA Reloaded, ~~which is crashing with BepInEx 6 for other reasons (as far as I can tell)~~, and Thomas Was Alone. @bbepis has tested this successfully with Risk of Rain 2.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

* [x] Bootstrap BepInEx.Unity with Doorstop 4
  * [x] Update CakeBuild script to download and bundle Doorstop 4
  * [x] Fix `ThreadingHelper` to not bootstrap on Unity 4
  * [x] ~~Use mono's native API for intercepting assembly loading events instead of `AppDomain.AssemblyLoad`~~ Instead use mono API to load assembly into memory with filename
* [x]  Bootstrap BepInEx.Il2Cpp with Doorstop 4
  * [x] Update Doorstop config
  * [x] Update CakeBuild script to bundle CoreCLR instead of mono BCL
  * [x] Upgrade TFM to net6.0
  * [x] Update Cpp2IL to latest version, remove AppDomain implementation
  * [x] Use dotnet's managed console encoder instead of custom WinAPI-based encoder used in UnityMono
  * [x] Integrate Il2CppInterop
  * [x] Integrate #409 and latest master
  * [x] ~~UnityDataMiner: start collecting mono corlibs for Unhollower use~~ not really needed due to how unhollower works
  * [x] Investigate crashes on Wine (added a temporary workaround for Il2CppInterop)

Other:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #562, opened 2023-03-03
Fix argument count when launched with Steam on Linux

zeroepoch:
## Description
Resolves a recent change (I think) with how Steam on Linux launches applications. I believe the extra `--` arguments are messing up the argument counts for the relaunch. Using the following change I was able to get the game working. Basically added 2 extra arguments and shifted the last argument by 2.

```
if [ "$2" = "SteamLaunch" ]; then
    "$1" "$2" "$3" "$4" "$5" "$6" "$0" "$7"
    exit
fi
```

## Motivation and Context
I had to resolve this issue locally in order to even launch the game on Steam. The changes I made to the game properties in Steam were following [this](https://docs.bepinex.dev/articles/advanced/steam_interop.html) link. This PR resolves issue #559.

## How Has This Been Tested?
I tested this change using the game Ultimate Chicken Horse on Steam on a laptop running Fedora 37.

## Types of changes

## Checklist:

### DIFF
--- a/doorstop/run_bepinex.sh
+++ b/doorstop/run_bepinex.sh
@@ -10,7 +10,7 @@ a="/$0"; a=${a%/*}; a=${a#/}; a=${a:-.}; BASEDIR=$(cd "$a"; pwd -P)
 # ---- EDIT AS NEEDED ------
 
 # EDIT THIS: The name of the executable to run
-# LINUX: This is the name of the Unity game executable 
+# LINUX: This is the name of the Unity game executable
 # MACOS: This is the name of the game app folder, including the .app suffix
 executable_name=""
 
@@ -32,7 +32,7 @@ export DOORSTOP_CORLIB_OVERRIDE_PATH=""
 # Special case: program is launched via Steam
 # In that case rerun the script via their bootstrapper to ensure Steam overlay works
 if [ "$2" = "SteamLaunch" ]; then
-    "$1" "$2" "$3" "$4" "$0" "$5"
+    "$1" "$2" "$3" "$4" "$5" "$6" "$0" "$7"
     exit
 fi
 
@@ -92,7 +92,7 @@ abs_path() {
 }
 
 _readlink() {
-    # relative links with readlink (without -f) do not preserve the path info 
+    # relative links with readlink (without -f) do not preserve the path info
     ab_path="$(abs_path "$1")"
     link="$(readlink "${ab_path}")"
     case $link in
@@ -105,8 +105,8 @@ _readlink() {
 
 resolve_executable_path () {
     e_path="$(abs_path "$1")"
-    
-    while [ -L "${e_path}" ]; do 
+
+    while [ -L "${e_path}" ]; do
         e_path=$(_readlink "${e_path}");
     done
     echo "${e_path}"


zeroepoch:
Any update on this? Is anyone else seeing the same issue with Steam games?

Ecconia:
Yeah I also had this issue on Linux when launching the BepInEx via Steam.
And your suggested fix works!
It also only makes sense given the current arguments that Steams provides via `%command%`.
Thanks for it.

For completeness sake, arguments in `./run_bepinex.sh %command%` are:
0: `./run_bepinex.sh` 1: `[...]/reaper` 2: `SteamLaunch` 3: `AppId=1234` 4: `--` 5: `[..]/steam-launch-wrapper` 6: `--` 7: `[..]/Game_Executable_By_Steam`

ManlyMarco (commented):
What happens if the user has an old version of Steam for whatever reason? Can it be made to work on both old and new version of the arguments?

zeroepoch:
> What happens if the user has an old version of Steam for whatever reason? Can it be made to work on both old and new version of the arguments? Also a comment explaining the arguments would be nice.

I don't think it will work with an older version of Steam. Since Steam is constantly being updated I don't know if you can go back to an older version.

ManlyMarco:
> I don't think it will work with an older version of Steam. Since Steam is constantly being updated I don't know if you can go back to an older version.

You can stay without updating in offline mode indefinitely as long as the game doesn't have network features or secondary DRM. Since you can also use BepInEx fully offline I think breaking backwards compatibility is not ideal.

If you have an old physical release with the full game files if I recall correctly you can install the Steam build from the disc and the game and it will work fine until you connect to the internet. This is not really relevant in this case though since I don't think there's any Unity games released in this format that we could support.

zeroepoch:
> It looks like you could detect how many arguments you have gotten and branch based on that?

I think that should be the case since the last argument `$7` should be non-empty in the newer case. TBH I don't have this setup anymore to test. Some friends and I were trying to get a game working together and since I spent an hour debugging this issue, and I didn't see anyone else report it, I didn't want the fix to be lost or the effort repeated by others.

ManlyMarco (approved):

doorstop/run_bepinex.sh
```diff
@@ -32,7 +32,7 @@ export DOORSTOP_CORLIB_OVERRIDE_PATH=""
 # Special case: program is launched via Steam
 # In that case rerun the script via their bootstrapper to ensure Steam overlay works
 if [ "$2" = "SteamLaunch" ]; then
-    "$1" "$2" "$3" "$4" "$0" "$5"
+    "$1" "$2" "$3" "$4" "$5" "$6" "$0" "$7"
```
astiob, line 35: @ManlyMarco I don’t have a Linux Steam to test this on, but if you just want to forward all arguments—that is, assuming that there aren’t additional arguments that Steam passes in but you want to drop—then this should be compatible with old and new versions alike:

```
count=0
until test $((count+=1)) -ge $#
do
    set -- "$@" "$1"
    shift
done
set -- "$@" "$0" "$1"
shift
"$@"
```

(Thanks to https://stackoverflow.com/a/62227833/865331 for a cross-platform implementation.)

Also, rather than `"$@"; exit` in this line and the line after it, you may want to `exec "$@"`.

Want me to submit a pull request?

ManlyMarco: > Want me to submit a pull request?

Yes, please do, that would be great. It'll have to be applied to the master branch as well.

astiob: Hmm, the version on master passes the remaining arguments through, as though they can be non-empty.  This originates in #391 without any apparent explanation. If extra arguments are really possible, then this should rather parse the command line, find the `steam-launch-wrapper [--]` and insert BepInEx after that, and it probably needs to preserve the extra arguments in v5, too. Alternatively, I guess it can just count the same fixed 5 arguments but avoid counting any instances of `--`.

PR #593, opened 2023-04-11
- Make BepInEx print the thunderstore manifest.json of a mod when available - DiskLogListener expose the file path that it uses

xiaoxiao921:
First change makes debugging much easier
Second change is needed for BepInExGUI purposes


xiaoxiao921:
PR on the wrong repo, my bad

PR #656, opened 2023-07-26
Make Logger.Listeners and Logger.Sources thread safe

### LINKED ISSUE #655: Logger.Listeners and Logger.Sources are not thread safe
### What happened?

Adding or removing `Logger.Listeners` and `Logger.Sources` can cause exceptions if some other thread is currently enumerating these lists, e.g. when calling `Logger.Log`. This usually causes unexpected crashes in whatever code was trying to log something at an unfortunate time.

### Steps to reproduce

1. Create a thread that adds and removes a Logger.Listeners in a loop.
2. Call `Logger.Log` on a plugin's logger in a loop.
3. Eventually a crash will happen.

It's possible to 100% get this to happen by trying to remove a listener from `Logger.Listeners` within a ILogListener.LogEvent, but that's a bit of a different issue.

### BepInEx Distribution

Stable from GitHub

### Log outputs

```txt
N/A
```

### Environment

```markdown
- OS: Any
- BepInEx: 5.4.21
- Game: Any
```

ManlyMarco:
Adding or removing `Logger.Listeners` and `Logger.Sources` can cause exceptions if some other thread is currently enumerating these lists, e.g. when calling `Logger.Log`. This usually causes unexpected crashes in whatever code was trying to log something at an unfortunate time. I noticed it can also lead to a listener not getting some log messages.

The changes result in sending log events being slightly faster with no allocations (saves a few ms overall), but Add/Remove operations are slightly slower (still 0 ms overall).

Fixes #655

## Types of changes

## Checklist:

### DIFF
--- a/BepInEx/Logging/Logger.cs
+++ b/BepInEx/Logging/Logger.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 
 namespace BepInEx.Logging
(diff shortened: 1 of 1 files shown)


ManlyMarco (commented):
After testing in a bunch of different games this seems to work perfectly fine so I'll just merge it in preparation for an upcoming BepInEx update.

PR #657, opened 2023-08-05
Not Download Unity Base Libraries If LatestUnityVersion==CurrentUnityVersion

Bright1192:
## Description

Not Download Unity Base Libraries If LatestUnityVersion==CurrentUnityVersion
## Motivation and Context

Sometime,when we may want to not download Unity Base Libraries the current Unity version when it is the same as the last Unity version
## Types of changes

## Checklist:

- [x] My code follows the code style of this project...Maybe?

### DIFF
--- a/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
+++ b/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
@@ -89,6 +89,8 @@ internal static partial class Il2CppInteropManager
 
     private static string HashPath => Path.Combine(IL2CPPInteropAssemblyPath, "assembly-hash.txt");
 
+    private static string LatestUnityVersionPath => Path.Combine(UnityBaseLibsDirectory, "latest-unity-version.txt");
+
     private static string UnityBaseLibsDirectory => Path.Combine(Paths.BepInExRootPath, "unity-libs");
 
     internal static string IL2CPPInteropAssemblyPath => Path.Combine(Paths.BepInExRootPath, "interop");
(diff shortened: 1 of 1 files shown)


Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
```diff
@@ -231,12 +228,29 @@ private static void GenerateInteropAssemblies()
...
+    private static bool CheckIfDownloadRequired()
+    {
+        if (!Directory.Exists(UnityBaseLibsDirectory))
+            return true;
+
+        if (!File.Exists(LatestUnityVersionPath))
+            return true;
+
+        var unityVersion = UnityInfo.Version;
+        if (new Version(unityVersion.Major, unityVersion.Minor, unityVersion.Build) != Version.Parse(File.ReadAllText(LatestUnityVersionPath)))
```
ManlyMarco, line 240: This can throw on malformed files.

Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
```diff
@@ -249,6 +263,7 @@ private static void DownloadUnityAssemblies()
             using var zipStream = httpClient.GetStreamAsync(source).GetAwaiter().GetResult();
             using var zipArchive = new ZipArchive(zipStream, ZipArchiveMode.Read);
 
+            File.WriteAllText(LatestUnityVersionPath, VERSION);
```
ManlyMarco, line 266: This should be written after all libraries successfully finish downloading and extracting.

Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
```diff
@@ -120,14 +122,9 @@ static void HashString(ICryptoTransform hash, string str)
...
 
         HashFile(md5, GameAssemblyPath);
-
-        if (Directory.Exists(UnityBaseLibsDirectory))
-            foreach (var file in Directory.EnumerateFiles(UnityBaseLibsDirectory, "*.dll",
-                                                          SearchOption.TopDirectoryOnly))
-            {
-                HashString(md5, Path.GetFileName(file));
-                HashFile(md5, file);
-            }
```
ManlyMarco, line 130: The hash check of base libraries should stay just in case they get corrupted for whatever reason.

ManlyMarco (changes requested):
Thank you for the PR! This looks like a good idea to me, it just needs some polish.

It looks like there's a merge conflict now so that would have to be resolved too.

PR #718, opened 2023-11-28
fix

DisrespectDwardo:
you're welcome


ManlyMarco (commented):
This is not a bug so there's nothing to fix. The console will stay on by default until RC and stable builds, where it will (most likely) be off by default.

PR #728, opened 2023-12-13
Add covariant interface for `ConfigEntry<T>`

Lordfirespeed:
## Description

I added an interfaces which isimplemented by `ConfigEntry<T>`, `ICovariantConfigEntry<T>`.

It'sapproximately as described - see [Microsoft's docs on variance in C#](https://learn.microsoft.com/en-us/dotnet/standard/generics/covariance-and-contravariance?redirectedfrom=MSDN#generic-delegates-with-variant-type-parameters). 

## Motivation and Context

The covariant interface allows retrieved values to be cast to a less derived type.
For example `ICovariantConfigEntry<*>` can be cast to `ICovariantConfigEntry<object>`,, which is useful if you want to add an event listener to the `SettingChanged` of many `ConfigEntry<T>` of varying `T`:
```cs
IEnumerable<ICovariantConfigEntry<object>> entries = new List<ICovariantConfigEntry<object>> 
{
    new ConfigEntry<bool>(...),
    new ConfigEntry<string>(...),
    new ConfigEntry<float>(...)
};

foreach (var entry in entries)
{
    entry.SettingChanged += foobar;
}
```

## How Has This Been Tested?

This has not been tested. 

## Types of changes

## Checklist:

### DIFF
--- a/BepInEx.Core/Configuration/ConfigEntryBase.cs
+++ b/BepInEx.Core/Configuration/ConfigEntryBase.cs
@@ -6,10 +6,27 @@ using BepInEx.Logging;
 namespace BepInEx.Configuration;
 
 /// <summary>
+///     Interface for covariant <see cref="Configuration.ConfigEntry&lt;T&gt;" />.
+/// </summary>
+/// <typeparam name="T">Type of the setting.</typeparam>
+public interface ICovariantConfigEntry<out T>
+{
+    /// <summary>
+    ///     Value of this setting. Get-only.
+    /// </summary>
+    public T Value { get; }
+    
+    /// <summary>
+    ///     Fired when the setting is changed. Does not detect changes made outside from this object.
+    /// </summary>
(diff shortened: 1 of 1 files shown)


ManlyMarco:
Looks like I made a mistake not placing the SettingChanged event inside the base class. Adding an interface just for this seems like an overkill to me, especially when the config system may still get changed before RC builds (in that case I'll just move the event to the base class). In the meantime you can subscribe to ConfigFile.SettingChanged to get updates on all config entries, and get their values through BoxedValue.

Lordfirespeed:
Yes, that is the workaround I've gone with for the time being! 

I'm glad to have put it on your radar. Would you like me to open an issue for 'moving SettingChanged event to ConfigBase'?

ManlyMarco:
Yeah, if you could open an issue for it it'd be great. I'll close this PR in that case.

PR #776, opened 2024-02-08
Add public GetInteropAssemblyPath

ManlyMarco:
Currently there is no public member exposing the interop assembly path. I'm not sure what the best place to put it would be. Preloader seems appropriate, either that or IL2CPPChainloader.

## Types of changes

## Checklist:


js6pak (changes requested):
1. This should be a property not a method
2. It shouldn't be in `Preloader` as it has no reason to be there. I don't really have a better idea though. `Il2CppInteropManager` would work, but we would have to make it public.

ManlyMarco:
> 2. It shouldn't be in `Preloader` as it has no reason to be there. I don't really have a better idea though. `Il2CppInteropManager` would work, but we would have to make it public.

How about repurposing `Paths.ManagedPath` for this? I believe it doesn't hold anything useful in IL2CPP currently and it's kind of the same thing.

ManlyMarco:
No longer required since #777 was merged

PR #946, opened 2024-08-07
Upgrade harmonyx to 2.13

aldelaro5:
This PR upgrades harmonyx to 2.13.

NOTE: This pr requires its matching pr for il2cppinterop which has small api changes made to work with this pr.

Draft for now due to some caveats about reorg monomod for the time being, but this is the bulk of the work to get it going so this pr can wait for now. I validated that it works for mono and il2cpp (with caveats, see below for details).

## Description

This does all the necessary prep work to upgrade harmonyx to 2.13 which means reorg MonoMod. However, there's currently 2 caveats related to MonoMod:
- It's not possible to make a CreateDetour implementation with the latest version of MonoMod. It's possible with CreateNativeDetour, but MonoMod doesn't allow me yet to modify the target method of the detour such that it would call the OrigEntrypoint. For now, it means that using MonoMod directly doesn't work for il2cpp, but since it didn't worked before (only harmonyx was supported), this isn't a regression so it's not a huge deal for now and shouldn't block this. I mention it to make it clear that the CreateDetour implementation is temporary (I put a TODO to make sure)
- MonoMod currently doesn't support Arm64. This however is a blocker because it is a regression which is the main reason I am submitting this as a draft for now in case more tests are desired. When monomod supports arm64, this could be ready for merge consideration.

Besides these 2 caveats, everything appears to work (see the test section below).

## Motivation and Context

It brings reorg monomod as well as a version of harmonyx with more features. Doing this upgrade also means that any future upgrade should be much more seamless from now on too.

## How Has This Been Tested?

I tested old and new mono on Windows, linux and macos (x64) and can confirm the following:

- Obra Dinn (old mono): Except for the AssetRipper.Primitives issue (see below for details), Windows and Linux works perfectly with UnityExplorer. Mac works with BepInEx, but it seems to not like that when it relaunches itself (due to steam drm), it...exits and bepinex doesn't work on the second launch. So BepInEx works, but it seems to break for an unrelated reason when relaunched.
- Marble It Up (new mono): Windows, Linux and Mac works perfectly with Unity explorer
- Among Us (il2cpp): I only tested Windows here, but detouring works including prefix / postfix with haromnyx. Again, monomod's Hook or IlHook direct usage can't work here for now.

Strangely, during testing, it was found that the only reason old mono doesn't boot as is is because it throws due to AssetRipper.Primitives using RegexOptions.Compiled. I will check with a set of corlibs from Unity (since the data miner doesn't yet publish old mono's corlibs), but either way, it's not related to this PR so it will need to be looked into separately.

## Types of changes

## Checklist:


BepInEx.Core/Console/Windows/WindowsConsoleDriver.cs
```diff
@@ -143,7 +144,7 @@ public void SetConsoleColor(ConsoleColor color)
 
     private static Stream OpenFileStream(IntPtr handle)
     {
-        if (ReflectionHelper.IsCore)
+        if (PlatformTriple.Current.Runtime.Target == RuntimeKind.CoreCLR)
```
nike4613, line 147: This can be more simply retrieved via `PlatformDetection.Runtime`

aldelaro5: done.

PR #960, opened 2024-08-20
Implement patcher and plugin providers

aldelaro5:
## Description

This pr implements the plugin provider and patcher provider features. It implicates many API changes, some are breaking to achieve this.

A plugin provider is a plugin, but its purpose isn't to modify the game, it is to provide the necessary facilities to load other plugins. The same goes for a patcher provider, but for patchers specifically. This implicates new sets of attributes and base classes as well as different implementation requirements. The main one is providing a `GetPlugins` / `GetPatchers` method that returns a list of `IPluginLoader`. This is called by the preloader or chainloader BEFORE actually loading any plugins.

An `IPluginLoader` is what allows BepInEx to load an assembly containing a plugin or patcher regardless of its source with optional caching support. It contains 3 important elements:

- An `AssemblyIdentifier` which the provider needs to provide which needs to be unique per assembly loaded by this patcher. This is used to track the assembly load among the others and for caching purposes
- An optional `AssemblyHash` which the provider can provide if it can as a caching optimisation. This field must change each time the assembly changes. If the provider can determine this without actually reading any data (for example, by having a separate hash), it allows the preloader or chainloader to quickly lookup the metadata information via caching without having to read or load the assembly. The main usecase is if it's determined that the plugin can't be loaded for whatever reasons, we avoid ANY read of the assembly. This is optional: a provider might leave this to null which effectively turns off caching
- A `GetAssemblyData` method which simply obtains the `byte[]` of the assembly, but doesn't yet load it into the app domain. This is what allows the provider to do whatever it wants to obtain the data and the chainloader to obtain the data when it is needed. Note however: the chainloader might call this multiple time so if this requires IO operations, the provider is responsible for lazy loading this for further calls. If `AssemblyHash` is provided however and the metadata were cached, the provider has the guarantee that the method won't be called unless the plugin is ready to be loaded

Providers optionally can have a custom assembly resolver. This is because since providers can supply plugins or patchers from any sources, their assembly resolutions needs to be custom too.

Additionally, for dogfooding purposes and also to give more liberty to custom distributions of BepInEx, both the default plugins and patchers loading behaviors were moved out of the chainloader / preloader and into their own independant providers which is shipped with BepInEx. This means the default loading behaviors could be disabled entirely. Effectively, the only plugin loading behavior that remains hardcoded is loading the integrated ones from the core folder, but BepInEx.Core and BepInEx.Core.Preloader no longer does anything with the plugins or patchers folder by themselves.

Speaking of folders, in order to better differentiate between providers and actual plugins, 2 new folders are introduced: `plugins-providers` and `patchers-providers`. Those do get examined by BepInEx and are fixed.

Finally, for internal tracking purposes, a `Loader` field is introduced to the corresponding metadata types which allows to track all this stuff for regular plugins. Providers leaves this to null and instead, they will be tracked using the `Location` field which serves the same purpose as before.

A few sidenotes:
- I noticed we overexpose a lot of the related apis. I decided to not fix them for now and do so in a further PR
- I am unsure about the way I modeled the attributes. If there's ways to make it better, let me know.
- There was a lot of debates over the type of `AssemblyHash` because being a string seems to make more sense from an api standpoint, but if this is a digest (such as SHA hashes), this carries a small overhead due to the type conversions. However, I decided that I am more in favor of this being a string: until I have real reasons to believe this optimisation is truly needed, I don't like the idea of making it `byte[]`.

## Motivation and Context

This feature was identified as a blocker for v6 being stable. It allows custom loaders or others to customise BepInEx in such a way that they can load plugins in any manner they want. For example: it is now possible to directly load plugins from a zip file or from a network...or anything really. Mod loaders can leverage this api to make it easier for them and it allows game independant loaders to exists.

## How Has This Been Tested?

I only tested on Unity (new mono) so far with Unity Explorer and a test patcher I made for Bug Fables. I didn't noticed anything breaking and could also load UE from a compressed zip file with a custom resolver working properly.

I have yet to test IL2CPP and FNA games.

## Types of changes

## Checklist:

### DIFF
--- a/BepInEx.Core/BepInEx.Core.csproj
+++ b/BepInEx.Core/BepInEx.Core.csproj
@@ -12,9 +12,6 @@
         <PackageReference Include="SemanticVersioning" Version="2.0.2"/>
         <PackageReference Include="MonoMod.Utils" Version="22.5.1.1"/>
     </ItemGroup>
-    <ItemGroup>
-        <Compile Remove="Contract\IPlugin.cs"/>
-    </ItemGroup>
 
     <!-- CopyLocalLockFileAssemblies causes to also output shared assemblies: https://github.com/NuGet/Home/issues/4837#issuecomment-354536302 -->
     <!-- Since all core assemblies usually follow naming of System.*, we just delete them for now -->
--- a/BepInEx.Core/BepInEx.Core.csproj
+++ b/BepInEx.Core/BepInEx.Core.csproj
@@ -12,6 +12,9 @@
         <PackageReference Include="SemanticVersioning" Version="2.0.2"/>
         <PackageReference Include="MonoMod.Utils" Version="22.5.1.1"/>
     </ItemGroup>
+    <ItemGroup>
+        <Compile Remove="Contract\IPlugin.cs"/>
+    </ItemGroup>
 
     <!-- CopyLocalLockFileAssemblies causes to also output shared assemblies: https://github.com/NuGet/Home/issues/4837#issuecomment-354536302 -->
     <!-- Since all core assemblies usually follow naming of System.*, we just delete them for now -->
--- a/BepInEx.Core/Bootstrap/BaseChainloader.cs
+++ b/BepInEx.Core/Bootstrap/BaseChainloader.cs
@@ -91,7 +91,7 @@ public abstract class BaseChainloader<TPlugin>
             Incompatibilities = incompatibilities,
             TypeName = type.FullName,
             TargettedBepInExVersion = bepinVersion,
-            Location = assemblyLocation,
+            _location = assemblyLocation,
             LoadContext = loadContext
         };
     }
@@ -390,10 +390,10 @@ public abstract class BaseChainloader<TPlugin>
                 Logger.Log(LogLevel.Info, $"Loading [{pluginInfo}]");
 
                 Assembly ass = null;
-                if (provider && !loadedAssemblies.TryGetValue(pluginInfo.Location, out ass))
+                if (provider && !loadedAssemblies.TryGetValue(pluginInfo._location, out ass))
                 {
-                    ass = Assembly.LoadFrom(pluginInfo.Location);
-                    loadedAssemblies[pluginInfo.Location] = ass;
+                    ass = Assembly.LoadFrom(pluginInfo._location);
+                    loadedAssemblies[pluginInfo._location] = ass;
                 }
                 else if (!provider && !loadedAssemblies.TryGetValue(pluginInfo.LoadContext.AssemblyIdentifier, out ass))
                 {
--- a/BepInEx.Core/Bootstrap/BaseChainloader.cs
+++ b/BepInEx.Core/Bootstrap/BaseChainloader.cs
@@ -403,20 +403,22 @@ public abstract class BaseChainloader<TPlugin>
                     var type = ass.GetType(plugin.TypeName);
                     plugin.Instance = Activator.CreateInstance(type);
                     AppDomain.CurrentDomain.AssemblyResolve += (_, args) => ((BasePluginProvider)plugin.Instance).ResolveAssembly(args.Name);
-                    var requestedPluginsStream = ((BasePluginProvider)plugin.Instance).GetPlugins();
-                    LoadContexts.AddRange(requestedPluginsStream);
+                    var pluginLoadContexts = ((BasePluginProvider)plugin.Instance).GetPlugins();
+                    LoadContexts.AddRange(pluginLoadContexts);
                     ProviderLoaded?.Invoke(plugin);
                 }
                 else
                 {
                     plugin.Instance = LoadPlugin(plugin, ass);
                     PluginLoaded?.Invoke(plugin);
+                    plugin.LoadContext.Dispose();
                 }
             }
             catch (Exception ex)
             {
                 invalidPlugins.Add(plugin.Metadata.GUID);
                 existingPlugins.Remove(plugin.Metadata.GUID);
+                plugin.LoadContext.Dispose();
 
                 Logger.Log(LogLevel.Error,
                            $"Error loading [{plugin}]: {(ex is ReflectionTypeLoadException re ? TypeLoader.TypeLoadExceptionToString(re) : ex.ToString())}");
--- a/BepInEx.Core/Bootstrap/BaseChainloader.cs
+++ b/BepInEx.Core/Bootstrap/BaseChainloader.cs
@@ -257,13 +257,12 @@ public abstract class BaseChainloader<TPlugin>
     /// <remarks>Some plugins may be skipped if they cannot be loaded (wrong metadata, etc).</remarks>
     /// <param name="plugins">Plugins to process.</param>
     /// <returns>List of plugins to load in the correct load order.</returns>
-    protected virtual List<T> ModifyLoadOrder<T>(IList<T> plugins, Dictionary<string, T> existingPlugins)
-        where T : PluginInfo
+    protected virtual List<PluginInfo> ModifyLoadOrder(IList<PluginInfo> plugins, Dictionary<string, PluginInfo> existingPlugins)
     {
         // We use a sorted dictionary to ensure consistent load order
         var dependencyDict =
             new SortedDictionary<string, IEnumerable<string>>(StringComparer.InvariantCultureIgnoreCase);
-        var pluginsByGuid = new Dictionary<string, T>();
+        var pluginsByGuid = new Dictionary<string, PluginInfo>();
 
         foreach (var pluginInfoGroup in plugins.GroupBy(info => info.Metadata.GUID))
         {
@@ -441,7 +440,7 @@ public abstract class BaseChainloader<TPlugin>
             {
                 invalidPlugins.Add(pluginInfo.Metadata.GUID);
                 existingPluginInfos.Remove(pluginInfo.Metadata.GUID);
-                pluginInfo.LoadContext.Dispose();
+                pluginInfo.LoadContext?.Dispose();
 
                 Logger.Log(LogLevel.Error,
                            $"Error loading [{pluginInfo}]: {(ex is ReflectionTypeLoadException re ? TypeLoader.TypeLoadExceptionToString(re) : ex.ToString())}");
--- a/BepInEx.Core/Bootstrap/BaseChainloader.cs
+++ b/BepInEx.Core/Bootstrap/BaseChainloader.cs
@@ -120,10 +120,12 @@ public abstract class BaseChainloader<TPlugin>
     ///     Obtains the <see cref="PluginInfo"/> corresponding to its GUID
     /// </summary>
     /// <param name="pluginGuid">The GUID of the plugin to get the info from</param>
-    /// <returns>The info of the plugin, null if the plugin wasn't loaded by the chainloader</returns>
-    public static PluginInfo GetPluginInfoFromGuid(string pluginGuid)
+    /// <param name="foundPluginInfo">When this method returns, contains the <see cref="PluginInfo"/>
+    /// associated with the specified GUID, if the plugin was loaded; otherwise, null. This parameter is passed uninitialized.</param>
+    /// <returns>true if the plugin with the GUID was loaded by the chainloader; otherwise, false.</returns>
+    public bool TryGetPluginInfoFromGuid(string pluginGuid, out PluginInfo foundPluginInfo)
     {
-        return Plugins.TryGetValue(pluginGuid, out var pluginInfo) ? pluginInfo : null;
+        return Plugins.TryGetValue(pluginGuid, out foundPluginInfo);
     }
 
     #region Contract
@@ -133,11 +135,16 @@ public abstract class BaseChainloader<TPlugin>
     private bool _initialized;
 
     /// <summary>
+    /// The current chainloader instance
+    /// </summary>
+    public static BaseChainloader<TPlugin> Instance { get; protected set; }
+    
+    /// <summary>
     ///     List of all <see cref="PluginInfo" /> instances of all plugins loaded via the chainloader.
     ///     If a plugin fails to load, it is removed. This list only contains the plugins that were actually
     ///     loaded and have their dependencies satisfied.
     /// </summary>
-    public static Dictionary<string, PluginInfo> Plugins { get; } = new();
+    public Dictionary<string, PluginInfo> Plugins { get; } = new();
 
     /// <summary>
     ///     List of all <see cref="PluginInfo" /> instances of all providers loaded via the chainloader.
@@ -147,8 +154,9 @@ public abstract class BaseChainloader<TPlugin>
     public Dictionary<string, PluginInfo> Providers { get; } = new();
     
     /// <summary>
-    ///     List of all <see cref="IPluginLoadContext" /> discovered from all the providers
-    ///     with each context associated with its provider info
+    ///     List of all <see cref="IPluginLoadContext" /> objects received from plugin providers.
+    ///     Value holds info of the originating provider. These contexts are used by chainloader to
+    ///     load plugins. Plugins that are loaded successfully are added to <see cref="Plugins" />.
     /// </summary>
     public Dictionary<IPluginLoadContext, PluginInfo> LoadContexts { get; } = new();
     
@@ -349,7 +357,7 @@ public abstract class BaseChainloader<TPlugin>
             var builtinProviders = DiscoverPluginProvidersFrom(Paths.BepInExAssemblyDirectory);
             var providers = DiscoverPluginProvidersFrom(Paths.PluginProviderPath);
             Logger.Log(LogLevel.Info, $"{providers.Count + builtinProviders.Count} " +
-                                      $"plugins provider{(providers.Count + builtinProviders.Count == 1 ? "" : "s")} to load");
+                                      $"plugin provider{(providers.Count + builtinProviders.Count == 1 ? "" : "s")} to load");
             
             LoadPlugins(builtinProviders, provider: true);
             LoadPlugins(providers, provider: true);
@@ -376,67 +384,67 @@ public abstract class BaseChainloader<TPlugin>
 
     private void LoadPlugins(List<PluginInfo> plugins, bool provider)
     {
-        var existingPlugins = provider ? Providers : Plugins;
-        var sortedPlugins = ModifyLoadOrder(plugins, existingPlugins);
+        var existingPluginInfos = provider ? Providers : Plugins;
+        var sortedPluginInfos = ModifyLoadOrder(plugins, existingPluginInfos);
         
         var invalidPlugins = new HashSet<string>();
         var processedPlugins = new Dictionary<string, SemanticVersioning.Version>();
         var loadedAssemblies = new Dictionary<string, Assembly>();
 
-        foreach (var plugin in sortedPlugins)
+        foreach (var pluginInfo in sortedPluginInfos)
         {
-            if (!CheckDependencies(plugin, processedPlugins, invalidPlugins, existingPlugins))
+            if (!CheckDependencies(pluginInfo, processedPlugins, invalidPlugins, existingPluginInfos))
             {
                 continue;
             }
 
             try
             {
-                Logger.Log(LogLevel.Info, $"Loading [{plugin}]");
+                Logger.Log(LogLevel.Info, $"Loading [{pluginInfo}]");
 
                 Assembly ass = null;
-                if (provider && !loadedAssemblies.TryGetValue(plugin.Location, out ass))
+                if (provider && !loadedAssemblies.TryGetValue(pluginInfo.Location, out ass))
                 {
-                    ass = Assembly.LoadFrom(plugin.Location);
-                    loadedAssemblies[plugin.Location] = ass;
+                    ass = Assembly.LoadFrom(pluginInfo.Location);
+                    loadedAssemblies[pluginInfo.Location] = ass;
                 }
-                else if (!provider && !loadedAssemblies.TryGetValue(plugin.LoadContext.AssemblyIdentifier, out ass))
+                else if (!provider && !loadedAssemblies.TryGetValue(pluginInfo.LoadContext.AssemblyIdentifier, out ass))
                 {
-                    ass = Assembly.Load(plugin.LoadContext.GetAssemblyData());
-                    loadedAssemblies[plugin.LoadContext.AssemblyIdentifier] = ass;
+                    ass = Assembly.Load(pluginInfo.LoadContext.GetAssemblyData());
+                    loadedAssemblies[pluginInfo.LoadContext.AssemblyIdentifier] = ass;
                 }
 
-                existingPlugins[plugin.Metadata.GUID] = plugin;
-                TryRunModuleCtor(plugin, ass);
+                existingPluginInfos[pluginInfo.Metadata.GUID] = pluginInfo;
+                TryRunModuleCtor(pluginInfo, ass);
 
                 if (provider)
                 {
-                    var type = ass.GetType(plugin.TypeName);
-                    plugin.Instance = Activator.CreateInstance(type);
-                    AppDomain.CurrentDomain.AssemblyResolve += (_, args) => ((BasePluginProvider)plugin.Instance).ResolveAssembly(args.Name);
-                    var pluginLoadContexts = ((BasePluginProvider)plugin.Instance).GetPlugins();
+                    var type = ass.GetType(pluginInfo.TypeName);
+                    pluginInfo.Instance = Activator.CreateInstance(type);
+                    AppDomain.CurrentDomain.AssemblyResolve += (_, args) => ((BasePluginProvider)pluginInfo.Instance).ResolveAssembly(args.Name);
+                    var pluginLoadContexts = ((BasePluginProvider)pluginInfo.Instance).GetPlugins();
                     foreach (IPluginLoadContext context in pluginLoadContexts)
                     {
-                        LoadContexts[context] = plugin;
+                        LoadContexts[context] = pluginInfo;
                     }
-                    ProviderLoaded?.Invoke(plugin);
+                    ProviderLoaded?.Invoke(pluginInfo);
                 }
                 else
                 {
-                    plugin.Source = LoadContexts[plugin.LoadContext];
-                    plugin.Instance = LoadPlugin(plugin, ass);
-                    PluginLoaded?.Invoke(plugin);
-                    plugin.LoadContext.Dispose();
+                    pluginInfo.Source = LoadContexts[pluginInfo.LoadContext];
+                    pluginInfo.Instance = LoadPlugin(pluginInfo, ass);
+                    PluginLoaded?.Invoke(pluginInfo);
+                    pluginInfo.LoadContext.Dispose();
                 }
             }
             catch (Exception ex)
             {
-                invalidPlugins.Add(plugin.Metadata.GUID);
-                existingPlugins.Remove(plugin.Metadata.GUID);
-                plugin.LoadContext.Dispose();
+                invalidPlugins.Add(pluginInfo.Metadata.GUID);
+                existingPluginInfos.Remove(pluginInfo.Metadata.GUID);
+                pluginInfo.LoadContext.Dispose();
 
                 Logger.Log(LogLevel.Error,
-                           $"Error loading [{plugin}]: {(ex is ReflectionTypeLoadException re ? TypeLoader.TypeLoadExceptionToString(re) : ex.ToString())}");
+                           $"Error loading [{pluginInfo}]: {(ex is ReflectionTypeLoadException re ? TypeLoader.TypeLoadExceptionToString(re) : ex.ToString())}");
             }
         }
     }
--- a/BepInEx.Core/Bootstrap/BaseChainloader.cs
+++ b/BepInEx.Core/Bootstrap/BaseChainloader.cs
@@ -19,6 +19,8 @@ public abstract class BaseChainloader<TPlugin>
 
     private static Regex allowedGuidRegex { get; } = new(@"^[a-zA-Z0-9\._\-]+$");
 
+    internal BaseChainloader() { }
+
     /// <summary>
     ///     Analyzes the given type definition and attempts to convert it to a valid <see cref="PluginInfo" />
     /// </summary>
@@ -369,6 +371,12 @@ public abstract class BaseChainloader<TPlugin>
         Logger.Log(LogLevel.Message, "Chainloader startup complete");
     }
 
+    [Obsolete("Use a plugin provider instead", true)]
+    public IList<PluginInfo> LoadPlugins(params string[] pluginsPaths)
+    {
+        throw new NotSupportedException("Use a plugin provider instead");
+    }
+    
     private void LoadPlugins(List<PluginInfo> plugins, bool provider)
     {
         var existingPluginInfos = provider ? Providers : Plugins;
--- a/BepInEx.Core/Bootstrap/BaseChainloader.cs
+++ b/BepInEx.Core/Bootstrap/BaseChainloader.cs
@@ -146,7 +146,7 @@ public abstract class BaseChainloader<TPlugin>
     ///     Value holds info of the originating provider. These contexts are used by chainloader to
     ///     load plugins. Plugins that are loaded successfully are added to <see cref="Plugins" />.
     /// </summary>
-    public Dictionary<IPluginLoadContext, PluginInfo> LoadContexts { get; } = new();
+    internal Dictionary<CachedPluginLoadContext, PluginInfo> LoadContexts { get; } = new();
     
     /// <summary>
     ///     Collection of error chainloader messages that occured during plugin loading.
@@ -350,7 +350,7 @@ public abstract class BaseChainloader<TPlugin>
             LoadPlugins(providers, provider: true);
             FinishedProviders?.Invoke();
             
-            var plugins = TypeLoader.GetPluginsFromLoadContexts(LoadContexts.Keys, ToPluginInfo<TPlugin>, HasBepinPluginType<BepInPlugin>, "chainloader");
+            var plugins = TypeLoader.GetPluginsFromLoadContexts(LoadContexts.Keys.Cast<IPluginLoadContext>(), ToPluginInfo<TPlugin>, HasBepinPluginType<BepInPlugin>, "chainloader");
             Logger.Log(LogLevel.Info, $"{plugins.Count} plugin{(plugins.Count == 1 ? "" : "s")} to load");
             LoadPlugins(plugins, provider: false);
             Finished?.Invoke();
@@ -414,23 +414,24 @@ public abstract class BaseChainloader<TPlugin>
                     var pluginLoadContexts = ((BasePluginProvider)pluginInfo.Instance).GetPlugins();
                     foreach (IPluginLoadContext context in pluginLoadContexts)
                     {
-                        LoadContexts[context] = pluginInfo;
+                        var cachedContext = new CachedPluginLoadContext(context);
+                        LoadContexts[cachedContext] = pluginInfo;
                     }
                     ProviderLoaded?.Invoke(pluginInfo);
                 }
                 else
                 {
-                    pluginInfo.Source = LoadContexts[pluginInfo.LoadContext];
+                    pluginInfo.Source = LoadContexts[(CachedPluginLoadContext)pluginInfo.LoadContext];
                     pluginInfo.Instance = LoadPlugin(pluginInfo, ass);
                     PluginLoaded?.Invoke(pluginInfo);
-                    pluginInfo.LoadContext.Dispose();
+                    foreach (CachedPluginLoadContext context in LoadContexts.Keys)
+                        context.Dispose();
                 }
             }
             catch (Exception ex)
             {
                 invalidPlugins.Add(pluginInfo.Metadata.GUID);
                 existingPluginInfos.Remove(pluginInfo.Metadata.GUID);
-                pluginInfo.LoadContext?.Dispose();
 
                 Logger.Log(LogLevel.Error,
                            $"Error loading [{pluginInfo}]: {(ex is ReflectionTypeLoadException re ? TypeLoader.TypeLoadExceptionToString(re) : ex.ToString())}");
--- a/BepInEx.Core/Bootstrap/BaseChainloader.cs
+++ b/BepInEx.Core/Bootstrap/BaseChainloader.cs
@@ -23,9 +23,10 @@ public abstract class BaseChainloader<TPlugin>
     ///     Analyzes the given type definition and attempts to convert it to a valid <see cref="PluginInfo" />
     /// </summary>
     /// <param name="type">Type definition to analyze.</param>
+    /// <param name="loadContext">The load context of the plugin</param>
     /// <param name="assemblyLocation">The filepath of the assembly, to keep as metadata.</param>
     /// <returns>If the type represent a valid plugin, returns a <see cref="PluginInfo" /> instance. Otherwise, return null.</returns>
-    public static PluginInfo ToPluginInfo<T>(TypeDefinition type, IPluginLoader loader, string assemblyLocation)
+    public static PluginInfo ToPluginInfo<T>(TypeDefinition type, IPluginLoadContext loadContext, string assemblyLocation)
     {
         if (type.IsInterface || type.IsAbstract)
             return null;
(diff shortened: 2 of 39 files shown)


BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -331,117 +347,127 @@ public virtual void Execute()
...
+                    plugin.Instance = Activator.CreateInstance(type);
+                    AppDomain.CurrentDomain.AssemblyResolve += (_, args) => ((BasePluginProvider)plugin.Instance).ResolveAssembly(args.Name);
+                    var requestedPluginsStream = ((BasePluginProvider)plugin.Instance).GetPlugins();
+                    Loaders.AddRange(requestedPluginsStream);
                 }
+                else
+                {
+                    plugin.Instance = LoadPlugin(plugin, ass);
+                }
+                PluginLoaded?.Invoke(plugin);
```
ManlyMarco, line 397: Should the event be called even for providers? Maybe it should include info that it is a provider or have a separate event?

BepInEx.Core/Contract/Attributes.cs
```diff
@@ -76,6 +76,32 @@ internal static BepInPlugin FromCecilType(TypeDefinition td)
...
+public class BepInPluginProvider : BepInPlugin
+{
+    /// <param name="GUID">The unique identifier of the plugin. Should not change between plugin versions.</param>
+    /// <param name="Name">The user friendly name of the plugin. Is able to be changed between versions.</param>
+    /// <param name="Version">The specific version of the plugin.</param>
+    public BepInPluginProvider(string GUID, string Name, string Version) : base(GUID, Name, Version)
+    {
+    }
+
+    internal static BepInPluginProvider FromCecilType(TypeDefinition td)
```
ManlyMarco, line 92:
```suggestion
    internal static new BepInPluginProvider FromCecilType(TypeDefinition td)
```

BepInEx.Core/Contract/PluginInfo.cs
```diff
@@ -16,6 +16,11 @@ public class PluginInfo : ICacheable
     ///     General metadata about a plugin.
     /// </summary>
     public BepInPlugin Metadata { get; internal set; }
+    
+    /// <summary>
+    ///     The loader used to load this plugin, null if it is a provider
+    /// </summary>
+    public IPluginLoader Loader { get; internal set; }
```
ManlyMarco, line 23: This could be null if the plugin was created later, e.g. by the ScriptEngine plugin whenever it gets ported to v6.

BepInEx.Core/Contract/PluginInfo.cs
```diff
@@ -33,24 +38,23 @@ public class PluginInfo : ICacheable
...
+    
     internal Version TargettedBepInExVersion { get; set; }
+    
+    internal string Location { get; set; }
+    
+    /// <inheritdoc />
+    public override string ToString() => $"{Metadata?.Name} {Metadata?.Version}";
 
-    void ICacheable.Save(BinaryWriter bw)
+    public virtual void Save(BinaryWriter bw)
```
ManlyMarco, line 55: Why change the caching methods to public? They shouldn't have any use outside of BepInEx.

BepInEx.Core/Contract/PluginInfo.cs
```diff
@@ -33,24 +38,23 @@ public class PluginInfo : ICacheable
     /// </summary>
     public IEnumerable<BepInIncompatibility> Incompatibilities { get; internal set; }
 
-    /// <summary>
-    ///     File path to the plugin DLL
-    /// </summary>
-    public string Location { get; internal set; }
```
ManlyMarco, line 39: There may be some trouble caused by this, some plugins relied on this to find their bundled resources. Not sure how to resolve this though, especially when loading from something like a zip file.

BepInEx.PatcherProvider/BepInExPatcherLoader.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.IO;
+
+namespace BepInEx.PatcherProvider;
+
+public class BepInExPatcherLoader : IPluginLoader
+{
+    public string AssemblyIdentifier { get; internal set; }
+    public string AssemblyHash { get; internal set; }
+    private byte[] assemblyData;
```
ManlyMarco, line 9: Holding the byte array forever is kind of suboptimal, especially when embedded resources are used. Maybe chainloader finishing could trigger all of these to get cleared? Or make this implement `IDisposable` and have chainloader dispose it after it will no longer need the ass data? This way would also force others to implement a Dispose method, forcing them to consider this issue I suppose.

BepInEx.Preloader.Core/Patching/AssemblyPatcher.cs
```diff
@@ -150,69 +157,155 @@ public void AddPatchersFromDirectory(string directory)
...
-                            sortedPatchers.Add(definition);
-                        }
+    private Assembly PatcherProvidersAssemblyResolver(object sender, ResolveEventArgs args)
+    {
+        foreach (BasePatcherProvider provider in providers)
+        {
+            var ass = provider.ResolveAssembly(args.Name);
+            if (ass != null)
+                return ass;
+        }
```
ManlyMarco, line 204: It might make sense to wrap the ResolveAssembly in a trycatch in case there's a single faulty provider so it doesn't bring down everything else with it,.

Runtimes/NET/BepInEx.NET.Common/BepInEx.NET.Common.csproj
```diff
@@ -12,6 +12,18 @@
...
+            <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+            <OutputItemType>Content</OutputItemType>
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+            <Targets>Build;DocumentationProjectOutputGroup</Targets>
+        </ProjectReference>
+        <ProjectReference Include="..\..\..\BepInEx.PatcherProvider\BepInEx.PatcherProvider.csproj">
+            <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+            <OutputItemType>Content</OutputItemType>
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+            <Targets>Build;DocumentationProjectOutputGroup</Targets>
```
ManlyMarco, line 25: Are these intended?

BepInEx.Core/Contract/IPluginLoader.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace BepInEx;
+
+/// <summary>
+///     An interface that allows to dynamically load and track an assembly
+///     which can be obtained from a provided
+/// </summary>
+public interface IPluginLoader
```
ManlyMarco, line 7: Maybe a "Source" property would be useful to track where the plugin came from?

ManlyMarco (commented):
Overall looks solid, I couldn't find any major issues. I mostly have comments about making the docs or API itself clearer.

BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -116,10 +122,20 @@ protected static bool PluginTargetsWrongBepin(PluginInfo pluginInfo)
...
 
+    /// <summary>
+    ///     List of all <see cref="PluginInfo" /> instances of all providers loaded via the chainloader.
+    /// </summary>
+    public Dictionary<string, PluginInfo> Providers { get; } = new();
+    
+    /// <summary>
+    ///     List of all <see cref="IPluginLoader" /> discovered from all the providers
+    /// </summary>
+    public List<IPluginLoader> Loaders { get; } = new();
```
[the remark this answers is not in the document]
aldelaro5: I thought about it and the reason I didn't is the method "GetAssemblyData" goes a bit beyond what I would characterise as an "info" container. It doesn't just provide info, but does the actual job of loading.

BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -306,15 +314,23 @@ protected virtual IList<PluginInfo> ModifyLoadOrder(IList<PluginInfo> plugins)
...
     /// </summary>
     public virtual void Execute()
     {
         try
         {
-            var plugins = DiscoverPlugins();
+            var builtinProviders = DiscoverPluginProvidersFrom(Paths.BepInExAssemblyDirectory);
+            var providers = DiscoverPluginProvidersFrom(Paths.PluginProviderPath);
+            Logger.Log(LogLevel.Info, $"{providers.Count + builtinProviders.Count} " +
+                                      $"meta plugin{(providers.Count + builtinProviders.Count == 1 ? "" : "s")} to load");
```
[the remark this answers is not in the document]
aldelaro5: That's a mistake: it was renamed during dev and I forgot to change it so will do.

BepInEx.Core/Contract/PluginInfo.cs
aldelaro5: I think this is a mistake because I previously messed a lot with the caching system, but those changes ended up not being necessary. Will recheck and revert if it is

Runtimes/NET/BepInEx.NET.Common/BepInEx.NET.Common.csproj
aldelaro5: Yes, it's to include the csproj, but not actually reference it so you still get it in the build output, but no one depends on it.

BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -25,14 +25,14 @@ public abstract class BaseChainloader<TPlugin>
     /// <param name="type">Type definition to analyze.</param>
     /// <param name="assemblyLocation">The filepath of the assembly, to keep as metadata.</param>
     /// <returns>If the type represent a valid plugin, returns a <see cref="PluginInfo" /> instance. Otherwise, return null.</returns>
-    public static PluginInfo ToPluginInfo(TypeDefinition type, string assemblyLocation)
+    public static PluginInfo ToPluginInfo<T>(TypeDefinition type, IPluginLoader loader, string assemblyLocation)
```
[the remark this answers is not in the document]
aldelaro5: There isn't. I would have to create one that each plugin type implements, but I concluded it could go in my pr that plans to cleanup the existing api anyway

BepInEx.Core/Contract/IPluginLoader.cs
aldelaro5: What would it contain? The PluginInfo of the provider?

ManlyMarco: Patcher's PluginInfo I guess? I could see it being useful in some cases, mostly debugging.

BepInEx.Core/Bootstrap/BaseChainloader.cs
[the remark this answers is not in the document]
ManlyMarco: If we went by what .NET does it would be PluginLoadContext I guess.

BepInEx.Core/Bootstrap/BaseChainloader.cs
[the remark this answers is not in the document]
ManlyMarco: Understandable.

BepInEx.Preloader.Core/Patching/AssemblyPatcher.cs
aldelaro5: Done

BepInEx.Core/Bootstrap/BaseChainloader.cs
aldelaro5: Added a separate event

BepInEx.Core/Bootstrap/BaseChainloader.cs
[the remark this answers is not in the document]
aldelaro5: I liked the name so I renamed it to IPluginLoadContext

BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -116,10 +122,20 @@ protected static bool PluginTargetsWrongBepin(PluginInfo pluginInfo)
...
     /// <summary>
-    ///     List of all <see cref="PluginInfo" /> instances loaded via the chainloader.
+    ///     List of all <see cref="PluginInfo" /> instances of all plugins loaded via the chainloader.
     /// </summary>
     public Dictionary<string, PluginInfo> Plugins { get; } = new();
 
+    /// <summary>
+    ///     List of all <see cref="PluginInfo" /> instances of all providers loaded via the chainloader.
+    /// </summary>
+    public Dictionary<string, PluginInfo> Providers { get; } = new();
```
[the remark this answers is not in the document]
aldelaro5: Done

BepInEx.PatcherProvider/BepInExPatcherLoader.cs
aldelaro5: Added IDosposable requirement and implemented it on the default 2 providers

BepInEx.Core/Contract/PluginInfo.cs
aldelaro5: As discussed, added a GetFile method that plugins can call

BepInEx.Core/Contract/IPluginLoader.cs
aldelaro5: Done, but I had to abstract all base plugins classes to an interface which strangely enough, said interface was already done, but its file was excluded from the solution.

Now, all 3 base plugins classes implements IPlugin which provides the stuff everyone was providing already, but with more consistency. The plugins still obtains their basic attribute metadata, but the rest is offloaded to the chainloader so the plugin can ask for their full info from their guid which will include the Source so all plugins can have all the info regardless of runtime including Source.

Alexejhero (approved):
I've been using this pr branch of BepInEx for a couple of days and it works well.

I haven't had any issues except breaking changes with other mods, since this PR makes public API changes without keeping the old contract with [Obsolete]. Is that normal for API changes in BepInEx 6?

BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -123,25 +134,40 @@ protected static bool PluginTargetsWrongBepin(PluginInfo pluginInfo)
 
     /// <summary>
     ///     List of all <see cref="PluginInfo" /> instances of all plugins loaded via the chainloader.
+    ///     If a plugin fails to load, it is removed. This list only contains the plugins that were actually
+    ///     loaded and have their dependencies satisfied.
     /// </summary>
-    public Dictionary<string, PluginInfo> Plugins { get; } = new();
+    public static Dictionary<string, PluginInfo> Plugins { get; } = new();
```
ManlyMarco, line 140: Why are Plugins the only static list?

BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -123,25 +134,40 @@ protected static bool PluginTargetsWrongBepin(PluginInfo pluginInfo)
...
     ///     List of all <see cref="PluginInfo" /> instances of all providers loaded via the chainloader.
+    ///     If a provider fails to load, it is removed. This list only contains the providers that were actually
+    ///     loaded and have their dependencies satisfied.
     /// </summary>
     public Dictionary<string, PluginInfo> Providers { get; } = new();
     
     /// <summary>
-    ///     List of all <see cref="IPluginLoader" /> discovered from all the providers
+    ///     List of all <see cref="IPluginLoadContext" /> discovered from all the providers
+    ///     with each context associated with its provider info
```
ManlyMarco, line 151:
```suggestion
    ///     List of all <see cref="IPluginLoadContext" /> objects received from plugin providers.
    ///     Value holds info of the originating provider. These contexts are used by chainloader to
    ///     load plugins. Plugins that are loaded successfully are added to <see cref="Plugins" >.
```

BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -323,12 +349,13 @@ public virtual void Execute()
             var builtinProviders = DiscoverPluginProvidersFrom(Paths.BepInExAssemblyDirectory);
             var providers = DiscoverPluginProvidersFrom(Paths.PluginProviderPath);
             Logger.Log(LogLevel.Info, $"{providers.Count + builtinProviders.Count} " +
-                                      $"meta plugin{(providers.Count + builtinProviders.Count == 1 ? "" : "s")} to load");
+                                      $"plugins provider{(providers.Count + builtinProviders.Count == 1 ? "" : "s")} to load");
```
ManlyMarco, line 352:
```suggestion
                                      $"plugin provider{(providers.Count + builtinProviders.Count == 1 ? "" : "s")} to load");
```

BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -387,19 +414,26 @@ private void LoadPlugins(List<PluginInfo> plugins, bool provider)
                     var type = ass.GetType(plugin.TypeName);
                     plugin.Instance = Activator.CreateInstance(type);
                     AppDomain.CurrentDomain.AssemblyResolve += (_, args) => ((BasePluginProvider)plugin.Instance).ResolveAssembly(args.Name);
-                    var requestedPluginsStream = ((BasePluginProvider)plugin.Instance).GetPlugins();
-                    Loaders.AddRange(requestedPluginsStream);
+                    var pluginLoadContexts = ((BasePluginProvider)plugin.Instance).GetPlugins();
+                    foreach (IPluginLoadContext context in pluginLoadContexts)
+                    {
+                        LoadContexts[context] = plugin;
```
ManlyMarco, line 420: Since the `plugin` local can now refer to either an actual plugin or a provider, I'd rename it to something like pluginInfo, same with other locals.

ManlyMarco (commented):
Looks good other than Plugins being made static. The rest is just improving docs.

BepInEx.PatcherProvider/BepInExPatcherProvider.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+using BepInEx.Logging;
+using BepInEx.Preloader.Core.Patching;
+
+namespace BepInEx.PatcherProvider;
+
+[PatcherProviderPluginInfo("io.bepinex.bepinexpatcherprovider", "BepInExPatcherProvider", "1.0")]
```
js6pak, line 10: I know `bepinex.io` is used as id in other places, but we don't own that domain, so I don't think we should use it. Just `BepInExPatcherProvider` would be fine probably.

BepInEx.Core/Contract/IPluginLoadContext.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// <summary>
+    ///     An identifier that uniquely identifies an assembly from a provider no matter its revision
+    /// </summary>
+    public string AssemblyIdentifier { get; }
+    
+    /// <summary>
+    ///     An optional hash that changes each time the assembly changes which can be tracked for cache
+    ///     invalidation purposes. If this is null, no caching occurs for this assembly load
+    /// </summary>
+    public string AssemblyHash { get; }
```
js6pak, line 20: Is AssemblyHash a proper name if it doesn't have to be a hash?

Runtimes/Unity/BepInEx.Unity.IL2CPP/BepInEx.Unity.IL2CPP.csproj
```diff
@@ -9,6 +9,18 @@
     </PropertyGroup>
     <ItemGroup>
         <ProjectReference Include="..\..\..\BepInEx.Core\BepInEx.Core.csproj"/>
+        <ProjectReference Include="..\..\..\BepInEx.PluginProvider\BepInEx.PluginProvider.csproj">
+            <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+            <OutputItemType>Content</OutputItemType>
```
js6pak, line 14: `Content` will flow to the consumers of the nuget package, we should use `None` instead

BepInEx.Preloader.Core/Patching/BasePatcherProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Reflection;
+using BepInEx.Configuration;
+using BepInEx.Logging;
+
+namespace BepInEx.Preloader.Core.Patching;
+
+/// <summary>
+///     A patcher provider that can discover and load assemblies containing patcher plugins
+/// </summary>
+public abstract class BasePatcherProvider
```
js6pak, line 11: This much code duplication with regular plugin providers is not acceptable imo

js6pak (changes requested):

BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -109,23 +116,66 @@
...
     }
 
+    /// <summary>
+    ///     Obtains the <see cref="PluginInfo"/> corresponding to its GUID
+    /// </summary>
+    /// <param name="pluginGuid">The GUID of the plugin to get the info from</param>
+    /// <param name="foundPluginInfo">When this method returns, contains the <see cref="PluginInfo"/>
+    /// associated with the specified GUID, if the plugin was loaded; otherwise, null. This parameter is passed uninitialized.</param>
+    /// <returns>true if the plugin with the GUID was loaded by the chainloader; otherwise, false.</returns>
+    public bool TryGetPluginInfoFromGuid(string pluginGuid, out PluginInfo foundPluginInfo)
```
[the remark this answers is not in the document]
aldelaro5: Because after discussion with manly, it was decided that in an api cleanup pr, `plugins` won't be public anymore, but rather, a read only collection of the plugins will be. This method could still be used to lookup from GUID.

BepInEx.Core/Bootstrap/BaseChainloader.cs
```diff
@@ -194,38 +244,30 @@ protected virtual void InitializeLoggers()
...
-    }
-
     /// <summary>
     /// Preprocess the plugins and modify the load order.
     /// </summary>
     /// <remarks>Some plugins may be skipped if they cannot be loaded (wrong metadata, etc).</remarks>
     /// <param name="plugins">Plugins to process.</param>
     /// <returns>List of plugins to load in the correct load order.</returns>
-    protected virtual IList<PluginInfo> ModifyLoadOrder(IList<PluginInfo> plugins)
+    protected virtual List<T> ModifyLoadOrder<T>(IList<T> plugins, Dictionary<string, T> existingPlugins)
```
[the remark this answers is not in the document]
aldelaro5: Seems it was needed in a previous version, but is no longer needed, good catch!

BepInEx.Core/Bootstrap/BaseChainloader.cs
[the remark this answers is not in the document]
js6pak: Isn't `IReadOnlyCollection<PluginInfo>` + `TryGetPluginInfoFromGuid` literally the same api surface as `IReadOnlyDictionary<string, PluginInfo>`?

BepInEx.Core/Contract/IPluginLoadContext.cs
aldelaro5: After discussion, it was determined it's fine for it to stay as is.

BepInEx.Core/Bootstrap/BaseChainloader.cs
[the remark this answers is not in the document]
aldelaro5: After discussions, I decided for this pr to use the TryGet directly, but we will have `Plugins`'s type change to be a readonly version of the dictionnary through some polyfills (due to ReadOnlyDictionnary not being avilable in net35)

aldelaro5:
I added mdb/ppdb symbols resolution in the api so providers can now resolve symbols of their assemblies

BepInEx.Core/Utility.cs
```diff
@@ -221,6 +221,55 @@ public static bool TryResolveDllAssembly<T>(AssemblyName assemblyName,
...
+    public static bool TryResolveDllSymbols(string dllPath, out byte[] symbolsData)
+    {
+        symbolsData = null;
+        
+        var subDirectory = Path.GetDirectoryName(dllPath);
+        var filename = Path.GetFileNameWithoutExtension(dllPath);
+
+        var potentialSymbolsFileNames = new[]
+        {
+            $"{filename}.mdb",
```
js6pak, line 239: `.mdb` is actually a suffix to the existing full file name like so `AssemblyName.dll.mdb`

BepInEx.PatcherProvider/BepInExPatcherLoadContext.cs
```diff
@@ -8,6 +8,7 @@ public class BepInExPatcherLoadContext : IPluginLoadContext
     public string AssemblyIdentifier { get; internal set; }
     public string AssemblyHash { get; internal set; }
     private byte[] assemblyData;
+    private byte[] assemblySymbolsData;
```
js6pak, line 11: You are not clearing this out in dispose

js6pak (changes requested):

BepInEx.Preloader.Core/Patching/BasePatcherProvider.cs
aldelaro5: I investigated and found that the best I can do is to extract an interface, but not remove the duplication until the second pr which will clean the api. The reason is I need to unity the plugins metadata types before I can do this which starts to be a bit out of scope for this pr

BepInEx.Core/Bootstrap/TypeLoader.cs
```diff
@@ -10,6 +10,47 @@
...
+    }
+
+    public byte[] GetFile(string relativePath)
+    {
+        if (!files.ContainsKey(relativePath))
+        {
+            files[relativePath] = PluginLoadContext.GetFile(relativePath);
+        }
+        return files[relativePath];
+    }
```
ManlyMarco, line 44: This might turn into a memory hog if plugins try to open large asset files this way. Also if the paths are case-insensitive this could hold multiple copies of the same file. Maybe it should only cache files below some size?

aldelaro5: I was heavily debating on having files caching in the first place, but I hadn't considered these cases so I removed it as plugins could decide themselves to do it if they really want it.

BepInEx.Core/Bootstrap/TypeLoader.cs
```diff
@@ -126,9 +163,8 @@ public static AssemblyDefinition CecilResolveOnFailure(object sender, AssemblyNa
     ///     selector.
     /// </returns>
     public static Dictionary<string, List<T>> FindPluginTypes<T>(string directory,
```
Alexejhero, line 165:
```
[Fatal  : Preloader] System.IO.DirectoryNotFoundException: Could not find a part of the path 'BepInEx\patchers-providers'.
```
You should probably make the folder if it doesn't exist

Alexejhero (changes requested):
See above

BepInEx.Core/Bootstrap/TypeLoader.cs
aldelaro5: Done, interestingly, it seems v6 wasn't correctly enforcing the patchers one (it was, but in the chainloader when the folder isn't used anymore by this point)

Alexejhero:
One last thing, I think this PR should not make any breaking changes. Updating BepInEx will be a hassle for modding communities if all mods would break and would have to be updated simultaneously. Ideally there would remain Obsolete implementations for all old methods/properties.

aldelaro5:
> One last thing, I think this PR should not make any breaking changes. Updating BepInEx will be a hassle for modding communities if all mods would break and would have to be updated simultaneously. Ideally there would remain Obsolete implementations for all old methods/properties.

This is why BE 2 was released: this pr already has breaking changes, but the reality is there's actually A LOT more planned because during the making of this pr, it revealed the entire plugins loading api desperately need a lot of cleanup. It's worth to do them now because not only v6 was never declared stable, but v6's goal is to remain stable for a long time. There's been many discussions about the current api not being good enough (overexposition and api surface duplication are the 2 big ones that came up) and a lot of these issues becomes much more apparent when I end up opening a pr that has to deal with a lot of them.

Essentially, v6 isn't currently stable so breaking changes like these are inevitable until it is stable.

Alexejhero:
You say that v6 isn't stable, but the truth is that modding communities such as Among Us use it as the de-facto mod loader and have done so for ages now. BepInEx v6 has been in bleeding edge at least since 2021 according to the oldest BE build, maybe even 2020 according to my memory. From an outsider perspective (to the BepInEx dev team), keeping it in BE for so long makes absolutely no sense given how well it works for normal usage. Continuing to hide behind its "unstable" status and making breaking changes can cause real harm to modding communities who have used this "unstable" version for years now.

Bundling a new feature (metaplugins) with breaking changes is a terrible idea, because if developers need the feature they will also be forced to also take the technical debt of these breaking changes, which will suddenly make older mods not work.

Additionally, the "breaking changes" in this specific PR are so minor they feel like nitpicks, such as renaming `PluginLoad` to `PluginLoaded` (and removing two of the parameters for whatever reason) and renaming `Log` to `Logger`. (At least those are the breaking changes that I ran into, ___for 3 different mods that I was depending on___. As such I am forced to use d8dd404e52d4669f54b665cb525f9b7119713a9e which is a commit that doesn't include those changes.)

Alexejhero:
Thank you for creating this feature, it is really useful to me.
That being said, if you do not wish to implement backwards compatibility, at least allow me to do it myself and pull request it into this branch, before it is merged into master.

aldelaro5:
> You say that v6 isn't stable, but the truth is that modding communities such as Among Us use it as the de-facto mod loader and have done so for ages now. BepInEx v6 has been in bleeding edge at least since 2021 according to the oldest BE build, maybe even 2020 according to my memory. From an outsider perspective (to the BepInEx dev team), keeping it in BE for so long makes absolutely no sense given how well it works for normal usage. Continuing to hide behind its "unstable" status and making breaking changes can cause real harm to modding communities who have used this "unstable" version for years now.
> 
> Bundling a new feature (metaplugins) with breaking changes is a terrible idea, because if developers need the feature they will also be forced to also take the technical debt of these breaking changes, which will suddenly make older mods not work.
> 
> Additionally, the "breaking changes" in this specific PR are so minor they feel like nitpicks, such as renaming `PluginLoad` to `PluginLoaded` (and removing two of the parameters for whatever reason) and renaming `Log` to `Logger`. (At least those are the breaking changes that I ran into, _**for 3 different mods that I was depending on**_. As such I am forced to use [d8dd404](https://github.com/BepInEx/BepInEx/commit/d8dd404e52d4669f54b665cb525f9b7119713a9e) which is a commit that doesn't include those changes.)

I have pushed 2 commits that I think solves your incompatibilities issues, but I need to explain how we (me and other bepinex maintainers) came to this decision because your comment sparkled a lively discussion about how we will version BepinEx moving forward.

Essentially, we found that "v6" as we know it has gone on for so long that people reasonably could perceive it as if v6 was stable, but obviously wasn't. This lengthy period can mostly be attributed to many key core devs either becoming less available to contribute or in the case of horse, completely MIA. This is something you can see for yourself if you check the commit history before and after January 2023 where horse went MIA.

The problem is eventually, we need to "rip the bandaid off" because it's not sustainable maintenance wise to release an LTS of v6 as it is. So we were into an impasse, but after much discussions, decided on the following:

- v6 will NEVER get a stable release. It will only have "pre" style release (there is going to be another release in this style that includes this PR WITH the commits I just pushed making it very minorly breaking, but with a different name to make it clear that it's the end of v6). The breaking changes we plan won't be included in this release, but it also will NOT be an LTS release. It means it's provided without much support because we do not declare it as a "v6.0"
- v7 will be the next LTS. This means we completely skip v6.0 and what we referred to as "stable v6" will now be referred to as v7.0. This is done to mark a very clear delineation where v7.0 is completely breaking from any BE or preview releases of v6, but it still communicates that v6 was perceived as sort of stable due to how much longer than planned its development took. That way, older guides that instructed to install v6 will still be correct and if people want to plan a migration to v7, they could

With the commits I just pushed, things should at least be a lot easier to work with. It will allow you to use the providers feature in that upcoming v6 release, but allows us to still have breaking changes and release it as v7 to make it clear that the "v6 unstable era" is over and that one will be LTS.

Alexejhero (approved):

ManlyMarco (approved):
Looks good, this should be non-breaking now.

BepInEx.PatcherProvider/BepInExPatcherLoadContext.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.IO;
+
+namespace BepInEx.PatcherProvider;
+
+public class BepInExPatcherLoadContext : IPluginLoadContext
```
js6pak, line 6: These shouldn't be public because people will reference `BepInEx/core/*.dll`, *even when a nuget package exists and is better...*

js6pak (changes requested):

BepInEx.PatcherProvider/BepInExPatcherLoadContext.cs
aldelaro5: Changed to internal

BepInEx.Core/BepInEx.Core.csproj
```diff
@@ -12,9 +12,6 @@
         <PackageReference Include="SemanticVersioning" Version="2.0.2"/>
         <PackageReference Include="MonoMod.Utils" Version="22.5.1.1"/>
     </ItemGroup>
-    <ItemGroup>
-        <Compile Remove="Contract\IPlugin.cs"/>
-    </ItemGroup>
```
js6pak, line 17: Should this be in this PR after all?

BepInEx.Core/Contract/PluginInfo.cs
```diff
@@ -77,7 +90,6 @@ void ICacheable.Save(BinaryWriter bw)
     void ICacheable.Load(BinaryReader br)
     {
         TypeName = br.ReadString();
-        Location = br.ReadString();
```
js6pak, line 80: Sidenote: this (I think?) will crash on older caches and result in skipping the cache via the try-catch, but should we make this versioned so we can skip on cache format updates more nicely?

BepInEx.Core/Utility.cs
```diff
@@ -221,6 +221,54 @@ public static bool TryResolveDllAssembly<T>(AssemblyName assemblyName,
...
 
+    /// <summary>
+    ///     Try to resolve the symbols data from an assembly DLL.
+    /// </summary>
+    /// <param name="dllPath">File path of the DLL assembly.</param>
+    /// <param name="symbolsData">The symbols of the DLL assembly.</param>
+    /// <returns>True, if the symbols were found. Otherwise, false.</returns>
+    public static bool TryResolveAssemblySymbols(string dllPath, out byte[] symbolsData)
+    {
+        symbolsData = null;
```
js6pak, line 232: nit (don't consider this blocking): assigning the out parameter in `TryGet*`  methods right before the `return` instead at the start will force you to think about the return value in all possible code paths

BepInEx.Core/Utility.cs
```diff
@@ -221,6 +221,54 @@ public static bool TryResolveDllAssembly<T>(AssemblyName assemblyName,
...
+        
+        var subDirectory = Path.GetDirectoryName(dllPath);
+
+        var potentialSymbolsFileNames = new[]
+        {
+            $"{dllPath}.mdb",
```
js6pak, line 238: Sidenote-ish:
Let's say a plugin wants to be universal and work (with debug symbols) on legacy mono, modern mono and coreclr platforms that have mdb, mdb/pdb and pdb support respectively.
Always preferring mdb over pdb will lead to: no symbols on coreclr, potentially worse symbols (if ppdb had more information than mdb, although I don't think that is the case today) on modern mono and only legacy mono would work as expected.

Reversing the order today would break legacy mono altogether so... I think it's fine as is, and we could reverse the order after adding the automatic debug symbols conversion that would figure out "the right thing" for the current runtime.

js6pak (changes requested):
This PR adds the following breaking changes:

1.
- `BepInEx.Bootstrap.BaseChainloader<TPlugin>.ToPluginInfo(Mono.Cecil.TypeDefinition, string)`
- `BepInEx.Bootstrap.BaseChainloader<TPlugin>.HasBepinPlugins(Mono.Cecil.AssemblyDefinition)`
- `BepInEx.Bootstrap.BaseChainloader<TPlugin>.DiscoverPluginsFrom(string, string)`
- `BepInEx.Bootstrap.BaseChainloader<TPlugin>.DiscoverPlugins()`
- `BepInEx.Bootstrap.BaseChainloader<TPlugin>.ModifyLoadOrder(System.Collections.Generic.IList<BepInEx.PluginInfo>)`
- `BepInEx.Bootstrap.TypeLoader.FindPluginTypes<T>(string, System.Func<Mono.Cecil.TypeDefinition, string, T>, System.Func<Mono.Cecil.AssemblyDefinition, bool>, string)`

All of the above are really only useful when you are implementing the `BaseChainloader`, which was never a supported use case? So we'll probably be fine with introducing those.
We are already [exposing `BepInEx.Core` internals to the runtimes](https://github.com/BepInEx/BepInEx/blob/6b38cee1277f4e3da5f2f685aee8b503a14d4054/BepInEx.Core/Properties/AssemblyInfo.cs) so we could make the `BaseChainloader` constructor internal to ban external use but I hope we can get rid of the runtime specific chainloaders in V7 anyway.

2.
- `BepInEx.Bootstrap.BaseChainloader<TPlugin>.LoadPlugins(params string[])`

This one could (and I've personally used it before) legitimately be used by (preloader) plugins as a crutch to load extra plugins without the plugin loader api.
There are likely close to zero mods using it so whether it's worth adding it back is questionable, especially because they can quite easily migrate from the hacky approach to plugin loaders.
We might want to add an`[Obsolete("use plugin loaders", true)]` + runtime exception though?

3.
- `BepInEx.Unity.Mono.Bootstrap.UnityChainloader.Instance`

This should probably get the same treatment as `IL2CPPChainloader.Instance`

BepInEx.Core/Utility.cs
aldelaro5: good idea, I ended up doing that

BepInEx.Core/BepInEx.Core.csproj
aldelaro5: With the compatibility accommodations, no so I reverted the changes

aldelaro5:
> This PR adds the following breaking changes:
> 
> 
>     * `BepInEx.Bootstrap.BaseChainloader<TPlugin>.ToPluginInfo(Mono.Cecil.TypeDefinition, string)`
> 
>     * `BepInEx.Bootstrap.BaseChainloader<TPlugin>.HasBepinPlugins(Mono.Cecil.AssemblyDefinition)`
> 
>     * `BepInEx.Bootstrap.BaseChainloader<TPlugin>.DiscoverPluginsFrom(string, string)`
> 
>     * `BepInEx.Bootstrap.BaseChainloader<TPlugin>.DiscoverPlugins()`
> 
>     * `BepInEx.Bootstrap.BaseChainloader<TPlugin>.ModifyLoadOrder(System.Collections.Generic.IList<BepInEx.PluginInfo>)`
> 
>     * `BepInEx.Bootstrap.TypeLoader.FindPluginTypes<T>(string, System.Func<Mono.Cecil.TypeDefinition, string, T>, System.Func<Mono.Cecil.AssemblyDefinition, bool>, string)`
> 
> 
> All of the above are really only useful when you are implementing the `BaseChainloader`, which was never a supported use case? So we'll probably be fine with introducing those. We are already [exposing `BepInEx.Core` internals to the runtimes](https://github.com/BepInEx/BepInEx/blob/6b38cee1277f4e3da5f2f685aee8b503a14d4054/BepInEx.Core/Properties/AssemblyInfo.cs) so we could make the `BaseChainloader` constructor internal to ban external use but I hope we can get rid of the runtime specific chainloaders in V7 anyway.
> 
> 
>     * `BepInEx.Bootstrap.BaseChainloader<TPlugin>.LoadPlugins(params string[])`
> 
> 
> This one could (and I've personally used it before) legitimately be used by (preloader) plugins as a crutch to load extra plugins without the plugin loader api. There are likely close to zero mods using it so whether it's worth adding it back is questionable, especially because they can quite easily migrate from the hacky approach to plugin loaders. We might want to add an`[Obsolete("use plugin loaders", true)]` + runtime exception though?
> 
> 
>     * `BepInEx.Unity.Mono.Bootstrap.UnityChainloader.Instance`
> 
> 
> This should probably get the same treatment as `IL2CPPChainloader.Instance`

I have done the 3 suggested changes here

BepInEx.Core/Contract/PluginInfo.cs
aldelaro5: I have added versioning at the CachedAssembly level so we avoid throwing in the case the cache changes

PR #974, opened 2024-09-01
global-metadata.dat file location fallback

MrPurple6411:
## Description
Quick fix for games that do not append program name to the Data folder..

## Motivation and Context
Games sometimes don't put the program name appended before _data in the file path.  I found it in [Voxtopolis](https://voxtopolis.com/)

## How Has This Been Tested?
I ran his game ... before change no work, after change works fine...

## Types of changes

## Checklist:
- [?] My code follows the code style of this project.


ManlyMarco (commented):
Rather than hardcoding another path, how about adding a path location setting? It would let you use a custom metadata file too.

PR #975, opened 2024-09-01
Fix failing to initialize if game's Data folder is just 'Data'

ManlyMarco:
Any Unity game can have its _Data folder renamed to just `Data` and continue working just fine. The issue is BepInEx assumes that there is always a game name prefix, which causes a startup crash.

Closes #974

## How Has This Been Tested?
Tested locally on a game with a bunch of plugins already installed. Renamed data folder to `Data` and tried to regenerate interop assemblies, which crashed on master but worked fine on this PR.

## Types of changes

### DIFF
--- a/BepInEx.Core/Paths.cs
+++ b/BepInEx.Core/Paths.cs
@@ -112,6 +112,9 @@ public static class Paths
         }
         else
         {
+            // According to some experiments, Unity checks whether globalgamemanagers/data.unity3d exists in the data folder before picking it.
+            // 'ProcessName_Data' folder is checked first, then if that fails 'Data' folder is checked. If neither is valid, the player crashes.
+            // A simple Directory.Exists check is accurate enough while being less likely to break in case these conditions change.
             GameDataPath = Path.Combine(GameRootPath, $"{ProcessName}_Data");
             if (!Directory.Exists(GameDataPath))
                 GameDataPath = Path.Combine(GameRootPath, "Data");
--- a/BepInEx.Core/Paths.cs
+++ b/BepInEx.Core/Paths.cs
@@ -104,8 +104,7 @@ public static class Paths
 
         GameRootPath = PlatformHelper.Is(Platform.MacOS)
                            ? Utility.ParentDirectory(executablePath, 4)
-                           : Path.GetDirectoryName(executablePath) 
-                          ?? throw new DirectoryNotFoundException("Failed to extract GameRootPath from executablePath: " + executablePath);
+                           : Path.GetDirectoryName(executablePath);
 
         if (managedPath != null && gameDataRelativeToManaged)
         {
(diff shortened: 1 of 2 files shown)


BepInEx.Core/Paths.cs
```diff
@@ -104,20 +104,31 @@ public static void SetExecutablePath(string executablePath,
 
         GameRootPath = PlatformHelper.Is(Platform.MacOS)
                            ? Utility.ParentDirectory(executablePath, 4)
-                           : Path.GetDirectoryName(executablePath);
+                           : Path.GetDirectoryName(executablePath) 
+                          ?? throw new DirectoryNotFoundException("Failed to extract GameRootPath from executablePath: " + executablePath);
```
js6pak, line 108: You are only applying this null check to `Path.GetDirectoryName` even though `Utility.ParentDirectory` will also return `null` in the same way. It will only return `null` for paths that don't have a parent like `/` or `C:/` so we are safe here though.
This should be tackled in a tree wide nullability PR not here.

BepInEx.Core/Paths.cs
```diff
@@ -104,20 +104,31 @@ public static void SetExecutablePath(string executablePath,
...
         ManagedPath = managedPath ?? Path.Combine(GameDataPath, "Managed");
         BepInExRootPath = bepinRootPath ?? Path.Combine(GameRootPath, "BepInEx");
         ConfigPath = Path.Combine(BepInExRootPath, "config");
         BepInExConfigPath = Path.Combine(ConfigPath, "BepInEx.cfg");
         PluginPath = Path.Combine(BepInExRootPath, "plugins");
         PatcherPluginPath = Path.Combine(BepInExRootPath, "patchers");
         BepInExAssemblyDirectory = Path.Combine(BepInExRootPath, "core");
-        BepInExAssemblyPath = Path.Combine(BepInExAssemblyDirectory,
-                                           $"{Assembly.GetExecutingAssembly().GetName().Name}.dll");
+        BepInExAssemblyPath = Path.Combine(BepInExAssemblyDirectory, $"{Assembly.GetExecutingAssembly().GetName().Name}.dll");
```
js6pak, line 131: Unrelated formatting change
Sidenote: why do we expose this

BepInEx.Core/Paths.cs
```diff
@@ -104,20 +104,31 @@ public static void SetExecutablePath(string executablePath,
 
         GameRootPath = PlatformHelper.Is(Platform.MacOS)
                            ? Utility.ParentDirectory(executablePath, 4)
-                           : Path.GetDirectoryName(executablePath);
+                           : Path.GetDirectoryName(executablePath) 
+                          ?? throw new DirectoryNotFoundException("Failed to extract GameRootPath from executablePath: " + executablePath);
+
+        if (managedPath != null && gameDataRelativeToManaged)
```
js6pak, line 110: How does `gameDataRelativeToManaged` work? When would it ever be the case?

js6pak (changes requested):

BepInEx.Core/Paths.cs
ManlyMarco: It's true in mono net35.

js6pak (approved):

BepInEx.Core/Paths.cs
```diff
@@ -106,9 +106,20 @@ public static void SetExecutablePath(string executablePath,
...
-        GameDataPath = managedPath != null && gameDataRelativeToManaged
-                           ? Path.GetDirectoryName(managedPath)
-                           : Path.Combine(GameRootPath, $"{ProcessName}_Data");
+        if (managedPath != null && gameDataRelativeToManaged)
+        {
+            GameDataPath = Path.GetDirectoryName(managedPath);
+        }
+        else
+        {
+            GameDataPath = Path.Combine(GameRootPath, $"{ProcessName}_Data");
```
ManlyMarco, line 115:
```suggestion
            // According to some experiments, Unity checks whether globalgamemanagers/data.unity3d exists in the data folder before picking it.
            // 'ProcessName_Data' folder is checked first, then if that fails 'Data' folder is checked. If neither is valid, the player crashes.
            // A simple Directory.Exists check is accurate enough while being less likely to break in case these conditions change.
            GameDataPath = Path.Combine(GameRootPath, $"{ProcessName}_Data");
```

PR #988, opened 2024-09-14
v7: Refactor the entire plugin loading and metadata API + improve documentation and access modifiers

aldelaro5:
## Description

This PR does a massive breaking api cleanup for v7 focused on 3 goals:

- Unify as much as possible the plugin loading and metadata API to reduce as much code duplication as possible
- Refine the access modifiers (except the detouring parts of IL2CPP since this area will get changed a lot later due to the planned reorg MonoMod upgrade)
- Improve xmldoc leaving no missing xml comments warning (same exception for the IL2CPP detouring part)

The second goal is mostly focused on restricting the api when things shouldn't be exposed and the third one is to solidify the documentation of what is exposed.

The first one is worth to elaborate on. The overall effect is the removal of 500-600 LOC thanks to code sharing as much as possible. That being said, this completely change the api in a very incompatible way, but it also simplifies it on the plugin developer side.

### Plugin classes
Essentially, there's no more runtime specific plugin class. There's only 2 now: GamePlugin and PatcherPlugin. Both of them inherit from a base class called Plugin which exists for the purpose of sharing and unifying the common properties all plugins have (Info, Logger and Config). As for the attributes, there's now only one to specify the familiar guid, name and version: BepInMetadata. All plugins must specify it no matter the type.

### Provider classes
As for providers, they work similarly in the sense there's a PatcherPluginProvider and GamePluginProvider, but either of these are empty class shells that inherit from PluginProvider which actually contains the api surface. This is done like that because it helps with assembly filtering and makes it clear on the developer side what kinds of plugin the provider will load. Since PluginProvider inherits from Plugin, they also share the same features plugins have.

### PluginInfo unification
Finally, PluginInfo is now used on every plugins no matter if it's a patcher, game plugin or provider. This allows every plugins to share the guid pool which allows patchers to gain features they didn't had access to before (mainly dependency resolution). It also allows plugins to depend on patchers since it becomes a lot easier to inform the chainloader of what was loaded already and it can leverage this information to resolve dependencies with what was loaded by the preloader.

### LoadingSystem
On the subject of loading, the entire loading logic was abstracted into a new class: LoadingSystem<TProvider, TPlugin>. The reasoning is that chainloaders and AssemblyPatcher were sharing a ton of loading logic that was duplicated with the main difference being the details of how the loading happened and which type of plugins the implementation was interested in. With this in mind, both were abstracted into LoadingSystem which now means that every plugins gets the same loading mechanism and dependency resolution.

Due to the amount of abstractions done, BaseChainloader (now renamed to Chainloader) is now able to share a lot more code such that the runtime specific portion of each chainloaders is very minimal. This allows to finally have a static chainloader instance that works accross runtimes because there's no longer anything exposed on the runtime specific chainloaders.

### Specific edgecases
Due to the unification, no plugins directly inherits from MonoBehavior. This introduces the need to be able to add components into the manager object, but 2 runtimes needs this exposed: unity mono and unity IL2CPP. This pr makes this work with a shared file accross the 2 csproj which exposes a method to add the component, but the implementation changes depending on the runtime via ifdef. This means the api still remains the exact same no matter which runtime is used, but the implementation can be adapted.

A similar file sharing trick is used for the default providers who have different logic, but needs to remain in 2 different assemblies.

Finally, UnityChainloader and its StaticStart can now be made internal because it was found that Unity 4.x's mono can work via an amend to the security attributes in the entrypoint patcher. Effectively, consumers can only see the Chainloader class and only specific members such as Plugins and Providers.

## Motivation and Context

It was very apparent that the plugin loading system had way too much duplicated code and wasn't overall ready to become more stable. This pr serves as a foundation to the v7 stability efforts.

This pr is mostly focused on refactoring the bulk of the api. Further items (such as enabling nullability) can be done after, but can be done once everyone agrees on this api refactor. In other words, this pr is the first step towards cleaning up things and further efforts will follow.

## How Has This Been Tested?

I been testing with a sample plugin + patcher using a unity 2018.4 game (new mono) and I also tested with a unity 4.5.5p5 game which also worked. I didn't observe regression there, but more testing should be done as I didn't had time to port more interesting plugins for this (this pr breaks the api so existing plugins would need to be adjusted to work with it).

What I am saying here is I highly recommend that further tests are done to make sure no regressions is introduced.

## Types of changes

## Checklist:


BepInEx.Core/Bootstrap/LoadingSystem.cs
```diff
@@ -1,34 +1,58 @@
...
+    /// </summary>
+    protected abstract string PluginLogName { get; }
+    
+    /// <summary>
+    ///     The name of the assembly load cache of this loading system
+    /// </summary>
+    protected abstract string LoadCacheName { get; }
 
-    internal BaseChainloader() { }
+    private static Regex AllowedGuidRegex => new(@"^[a-zA-Z0-9\._\-]+$");
```
ManlyMarco, line 44: This breaks every plugin, the general convention is to do `autor.pluginname` or just `pluginname` since this is what appears on dependency errors.

BepInEx.Core/Bootstrap/LoadingSystem.cs
```diff
@@ -154,7 +182,7 @@ protected static bool PluginTargetsWrongBepin(PluginInfo pluginInfo)
     ///     Collection of error chainloader messages that occured during plugin loading.
     ///     Contains information about what certain plugins were not loaded.
     /// </summary>
-    public List<string> DependencyErrors { get; } = new();
+    private List<string> DependencyErrors { get; } = new();
```
ManlyMarco, line 185: This needs to be public for https://github.com/BepInEx/BepInEx.Utility/blob/master/BepInEx.MessageCenter/MessageCenter.cs#L33-L34

BepInEx.Core/Bootstrap/LoadingSystem.cs
```diff
@@ -174,56 +202,7 @@ protected static bool PluginTargetsWrongBepin(PluginInfo pluginInfo)
...
-                                                     ConfigDiskAppend.Value, ConfigDiskLoggingInstantFlushing.Value,
-                                                     ConfigDiskLoggingFileLimit.Value));
-
-        if (!TraceLogSource.IsListening)
-            Logger.Sources.Add(TraceLogSource.CreateSource());
-
-        if (!Logger.Sources.Any(x => x is HarmonyLogSource))
-            Logger.Sources.Add(new HarmonyLogSource());
-    }
+    public event Action AllPluginsLoaded;
```
ManlyMarco, line 205: A progress event would be much more useful I think, with args saying what plugin is about to be loaded, was loaded, when loading started and then finished. Ideally the event would be static and fire for all LoadingSystems, just have the originating LoadingSystem in the args. It would make https://github.com/BepInEx/BepInEx.Debug/?tab=readme-ov-file#startup-profiler trivial.

BepInEx.Core/Configuration/ConfigFile.cs
```diff
@@ -46,21 +48,6 @@ public ConfigFile(string configPath, bool saveOnInit, BepInPlugin ownerMetadata)
 
     private Dictionary<ConfigDefinition, string> OrphanedEntries { get; } = new();
```
ManlyMarco, line 49: There needs to be a way to get these without Binding, it's been a fairly commonly requested feature for setting migration and dynamic settings.

BepInEx.Core/Contract/Attributes.cs
```diff
@@ -234,22 +206,22 @@ internal static IEnumerable<BepInIncompatibility> FromCecilType(TypeDefinition t
...
+        ProcessName = processName;
     }
 
     /// <summary>
     ///     The name of the process that this plugin will run under.
     /// </summary>
-    public string ProcessName { get; protected set; }
+    public string ProcessName { get; }
 
     internal static List<BepInProcess> FromCecilType(TypeDefinition td)
```
ManlyMarco, line 226: Since every attribute needs a FromCecilType maybe have all attributes inherit from an abstract class with this method in it as abstract?

BepInEx.Preloader.Core/Patching/AssemblyPatcher.cs
```diff
@@ -412,10 +167,69 @@ public static bool TryLoadAssembly(string path, out AssemblyDefinition assembly)
...
+                                              typeof(AssemblyDefinition).MakeByRefType()
+                                           && targetTypes.Length == 0
+                                           || parameters[0].ParameterType == typeof(TypeDefinition)
+                                           && targetAssemblies.Length == 0
+                                          )
+                                      || parameters.Length == 2 &&
+                                         parameters[1].ParameterType != typeof(string)
+                                      || method.ReturnType != typeof(void) &&
+                                         method.ReturnType != typeof(bool)
+               )
```
ManlyMarco, line 205: This is awful ;)

Runtimes/Unity/BepInEx.Unity.Common/ConfigUnityLog.cs
```diff
@@ -0,0 +1,16 @@
...
+internal static class ConfigUnityLog
+{
+    internal static readonly ConfigEntry<bool> ConfigUnityLogging = ConfigFile.CoreConfig.Bind(
+     "Logging", "UnityLogListening",
+     true,
+     "Enables showing unity log messages in the BepInEx logging system.");
+
+    internal static readonly ConfigEntry<bool> ConfigDiskWriteUnityLog = ConfigFile.CoreConfig.Bind(
+     "Logging.Disk", "WriteUnityLog",
+     false,
```
ManlyMarco, line 14: I think this should be on by default, many newer games don't even output PlayerLog and it's annoying to put the two logs together time-wise afterwards.

ManlyMarco (commented):
Overall good changes.

BepInEx.Core/Contract/Attributes.cs
aldelaro5: Overridable methods cannot be static. We also can't have a static member in an interface either so this is why we have to define this once per type

Runtimes/Unity/BepInEx.Unity.Common/ConfigUnityLog.cs
aldelaro5: I think the question is more why was it off by default in the first place?

I was thinking it had to do with games spamming this and if that was why...I don't see how that changes? The game would still have that problem if so with v7

BepInEx.Core/Bootstrap/LoadingSystem.cs
aldelaro5: Decided to make public for now, but am open to suggestions to improve on this

BepInEx.Core/Bootstrap/LoadingSystem.cs
aldelaro5: Wait, how's that possible?

All plugins and patchers were doing this on v6

ManlyMarco: Wait, my bad, I didn't look closely and thought it was an actual GUID regex. In this case it should be fine.

BepInEx.Core/Configuration/ConfigFile.cs
aldelaro5: reverted to public with an xmldoc

BepInEx.Preloader.Core/Patching/AssemblyPatcher.cs
aldelaro5: As discussed, a planned refactor that changes the patcher api will remove this code so it can temporarily stay for now

BepInEx.Core/Bootstrap/LoadingSystem.cs
aldelaro5: Changed events to use event args and be static, the progression part is something I am unsure, but it can be added later in case

Runtimes/Unity/BepInEx.Unity.Common/ConfigUnityLog.cs
aldelaro5: After discussion, I am fine with this so I changed it to true by default.

ManlyMarco (approved):

aldelaro5:
Marked as draft because I will be working on a superceding pr to this as other reviews are taking longer than I expected while I still want to refactor the api with a slightly different model than this.

I intended this to be an intermediary step, but it is not worth the waiting to get approval without planning.

aldelaro5:
Superceded by #999

PR #990, opened 2024-09-15
v6: Workaround the Assembly.Location being empty on v6

aldelaro5:
## Description

This workarounds a regression that was observed on v6 when a plugin would do Assembly.Location and get nothing because the plugin is now loaded in memory with a provider. Only the providers knows the actual location so as a workaround for v6, this pr postfixes Assembly.Location to return the correct path if applicable.

## Motivation and Context

This is a v6 regression and blocks its next release.

## How Has This Been Tested?

I tested from a unity 2018.4.12 game (new mono).

## Types of changes

## Checklist:

### DIFF
--- a/BepInEx.PatcherProvider/BepInExPatcherProvider.cs
+++ b/BepInEx.PatcherProvider/BepInExPatcherProvider.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Reflection;
 using BepInEx.Logging;
 using BepInEx.Preloader.Core.Patching;
+using HarmonyLib;
 
 namespace BepInEx.PatcherProvider;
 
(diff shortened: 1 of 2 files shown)


aldelaro5:
It was found that this pr breaks on xunity.autotranslator due to a legacy monomod issue, danike wrote:

summary:
1. coreclr vtables store either 1) pointers to the actual compiled method code, or 2) pointers to prestubs to that code
2. when asked for a methoddesc, it looks up that pointer, then maps that back to an actual methoddesc pointer a. for precode, this is done by looking at the specific known structure of the precode to find a location for a methoddesc pointer (simplified)
3. MonoMod is overwriting the precode with  a push; ret pair for an absolute jump, which breaks the precode recognition
4. runtime now doesn't know what methoddesc it corresponds to, causing the jit's devirt to fail

They point out that either it's very likely fixed in reorg or easy to fix if it wasn't. While this could be tested in v7 with the pending reorg upgrade pr, it means that it calls into question whether we can have providers in v6 because reorg is a massive breaking change which isn't a fit for v6 and...providers needs reorg to not introduce a rather major regression.

The end result is we're at an impasse and it seems we can't ship providers in v6. @ManlyMarco should I close this pr?

aldelaro5:
As discussed, I am closing the pr for the reasons mentioned above.

PR #994, opened 2024-09-18
v7: Change the chainloader Unity Mono entrypoint method to be Harmony based

aldelaro5:
## Description

This changes the way the chainloader entrypoint works on Unity mono to select one of 2 entrypoints that were proven to be quite reliable on il2cpp and relatively equivalent in timing with the default Application's cctor one. The big difference with this entrypoint however is it's possible games could pull the cctor earlier than expected while the one in this pr can't since it's a fixed point in the engine's boot process.

It was also found that the entrypoint could be acheived with a harmony patch applied as soon as possible (when the assembly gets loaded) and unapplied immediately once the entrypoint is reached. This provides a bunch of benefits including:

- We no longer need to ship a prepatcher which simplifies things
- It's a lot less intrusive which allows to, by default, no load UnityEngine.CoreModule in memory. This addresses an issue with the Rider's debugger evaluator not working correctly because it needs to have that module loaded from disk
- It simplifies the settings because this entrypoint is known to work and we no longer need to have users select their entrypoint

Additionally, it was found that StaticStart wasn't correctly protected against double calling. I fixed that in this PR.

Most of this was suggested and researched by @js6pak , but I validated that everything works (old or new mono, windows or linux) and improved the code they provided a bit.

## Motivation and Context

It simplifies the codebase, it simplifies the way we do the chainloader entrypoint and it simplifies settings. It also allows to have a similar entrypoint scheme for mono and il2cpp.

## How Has This Been Tested?

I tested on:
- Windows unity 2018.4 (new mono)
- Windows unity 2017.x (old mono)
- Windows unity 4.5.5 (old mono)
- Linux unity 2021.3 (new mono)

And validated that everything works still.

## Types of changes

## Checklist:


Runtimes/Unity/BepInEx.Unity.Mono.Preloader/UnityPreloader.cs
```diff
@@ -130,6 +137,61 @@ public static void Run()
...
+        var displayType = assembly.GetType(displayTypeName, false);
+        if (displayType != null)
+        {
+            var recreateDisplayListMethod = displayType.GetMethod("RecreateDisplayList", BindingFlags.NonPublic | BindingFlags.Static);
+            Harmony.Patch(recreateDisplayListMethod, postfix: new HarmonyMethod(typeof(UnityPreloader), nameof(Entrypoint)));
+            Log.LogInfo($"Hooked into {recreateDisplayListMethod.FullDescription()}");
+            return;
+        }
+
+        Log.LogError($"Couldn't find a suitable chainloader entrypoint in the {assemblyName} assembly because " +
```
ManlyMarco, line 173: This should be a Fatal log, also a trycatch wraparound would be a good idea to provide a better error message.

Runtimes/Unity/BepInEx.Unity.Mono.Preloader/UnityPreloader.cs
```diff
@@ -130,6 +137,61 @@ public static void Run()
...
+            Harmony.UnpatchSelf();
+
+            Assembly.Load("BepInEx.Unity.Mono")
+                    .GetType("BepInEx.Unity.Mono.Bootstrap.UnityChainloader")
+                    .GetMethod("StaticStart", AccessTools.all)!
+                    .Invoke(null, [null]);
+        }
+        catch (Exception e)
+        {
+            Log.LogError(e);
```
ManlyMarco, line 191: Should be fatal.

Runtimes/Unity/BepInEx.Unity.Mono.Preloader/UnityPreloader.cs
aldelaro5: Done and added a try catch with a different fatal error

ManlyMarco (approved):
Tested in various versions of 4.6, 5.6, 2017, works well.

Runtimes/Unity/BepInEx.Unity.Mono/Bootstrap/UnityChainloader.cs
```diff
@@ -71,6 +71,8 @@ public static void StaticStart(string gameExePath = null)
             if (staticStartHasBeenCalled)
                 throw new InvalidOperationException("Cannot call StaticStart again");
```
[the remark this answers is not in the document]
aldelaro5: Decided to log a fatal + return which is a bit safer.

PR #995, opened 2024-09-18
v6: Revert "Merge pull request #960 from aldelaro5/metaplugins"

aldelaro5:
This reverts commit 2a61b614978b1f3388c41f7784c43178ba98b58e, reversing changes made to 6b38cee1277f4e3da5f2f685aee8b503a14d4054.

## Description

This reverts the provider feature in v6 leaving it as a v7 only feature.

## Motivation and Context

Unfortunately, after thorough testing, we found that this feature introduces a breaking regression in v6 when accessing Assembly.Location. An attempt to workaround it was done in #990 , but it was found that this would run into a legacy monomod issue that is very likely fixed by upgrading to reorg monomod.

This means our only options are to leave this regression or to upgrade to reorg monomod. Neither are good: they each introduce breaking changes. It no longer becomes "easy" to bring this feature for v6 so this pr reverts it, leaving it in v7 only.

## How Has This Been Tested?

I checked the pr seemingly reverted correctly and tested on a unity 2018.4 (new mono) game to be sure and nothing broke.

## Types of changes

## Checklist:


ManlyMarco (approved):
All good.

PR #998, opened 2024-09-29
Make UnityInput and KeyboardShortcut available in IL2CPP

ManlyMarco:
## Description
Make UnityInput and KeyboardShortcut available in IL2CPP as well.

## Motivation and Context
With the merging of https://github.com/BepInEx/BepInEx/commit/d9b95a0ac76e53c4f0f2be93939adb02cfde16d4 the implication was that it would be available in IL2CPP as well, since that's the only version of v6 that matters. Also this was not the latest iteration of UnityInput.

~~This PR has an issue with references to the input dlls which needs to be resolved before merging. I don't know where the IL2CPP reference libs are coming from in the build process so I just used temporary files.~~

Since there's no nuget packages we can use for the input assemblies I just included a set of unhollowed assemblies. They can be replaced at a later point whenever there's proper packages on BepInEx nuget.

## How Has This Been Tested?
UnityInput was tested in many IL2CPP games by being included as-is in RuntimeUnityEditor.
I tested it after being included in BepInEx and it works the same way.

## Types of changes

## Checklist:


Runtimes/Unity/BepInEx.Unity.Mono/UnityInput.cs
```diff
@@ -29,18 +30,22 @@ public static IInputSystem Current
...
+                            var newInputSystem = new NewInputSystem();
+                            current = newInputSystem;
+                            BepInEx.Logging.Logger.Log(LogLevel.Debug, "[UnityInput] Using NewInputSystem");
+                        }
                     }
                     catch (Exception ex)
                     {
-                        Logging.Logger.Log(LogLevel.Warning, "[UnityInput] Failed to detect available input systems - " + ex);
+                        current = new NullInputSystem();
+                        BepInEx.Logging.Logger.Log(LogLevel.Debug, "[UnityInput] Failed to detect available input systems - " + ex);
```
ManlyMarco:
```suggestion
                        BepInEx.Logging.Logger.Log(LogLevel.Warning, "[UnityInput] Failed to detect available input systems - " + ex);
```

Runtimes/Unity/BepInEx.Unity.IL2CPP/UnityInput.cs
```diff
@@ -0,0 +1,898 @@
...
+                        {
+                            var newInputSystem = new NewInputSystem();
+                            current = newInputSystem;
+                            BepInEx.Logging.Logger.Log(LogLevel.Debug, "[UnityInput] Using NewInputSystem");
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        current = new NullInputSystem();
+                        BepInEx.Logging.Logger.Log(LogLevel.Debug, "[UnityInput] Failed to detect available input systems - " + ex);
```
ManlyMarco:
```suggestion
                        BepInEx.Logging.Logger.Log(LogLevel.Warning, "[UnityInput] Failed to detect available input systems - " + ex);
```

PR #1024, opened 2024-12-09
Checks if target Unity binaries archive exists before attempting download in case game's network conenction is blocked

lampuiho:
## Description
This allows downloading the archive manually to unity-libs folder so that if the game is blocked by firewall, it won't throw an exception.
Perhaps another solution is to prompt the user to download from the default web browser but for now this fixes the problem for me and is easy to code (coded this ages ago, I don't even remember when).

## Motivation and Context
I block all outbounds by default so naturally none of the games can connect to the internet.

## How Has This Been Tested?
fresh game folder with 2019.4.11.f Unity
download 2019.4.11.zip to `BepInEx\unity-libs`
copy files from `bin\Unity.IL2CPP` to `BepInEx\core`
copy, from release, the dotnet runtime and everything outside BepInEx to '.\', and dobby.dll to  `BepInEx\core`
run the game once, see the archive decompressed with no more HTTP exception

## Types of changes
- [x ] New feature (non-breaking change which adds functionality)

## Checklist:
- [x ] My code follows the code style of this project.
- [x ] My change requires a change to the documentation.
- [x ] I have updated the documentation accordingly.

### DIFF
--- a/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
+++ b/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
+using System.Linq;
 using System.Net.Http;
 using System.Reflection;
 using System.Security.Cryptography;
--- a/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
+++ b/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
@@ -287,6 +287,8 @@ internal static partial class Il2CppInteropManager
         var unityVersion = UnityInfo.Version;
         var version = $"{unityVersion.Major}.{unityVersion.Minor}.{unityVersion.Build}";
         var source = UnityBaseLibrariesSource.Value.Replace("{VERSION}", version);
+        if (string.IsNullOrEmpty(source)) return;
+
         var uri = new Uri(source);
         string file = Path.GetFileName(uri.AbsolutePath);
 
--- a/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
+++ b/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
@@ -282,25 +282,24 @@ internal static partial class Il2CppInteropManager
         }
     }
 
-    private static void DownloadUnityAssemblies()
-    {
+    private static void DownloadUnityAssemblies() {
         var unityVersion = UnityInfo.Version;
-        var source =
-            UnityBaseLibrariesSource.Value.Replace("{VERSION}",
-                                                   $"{unityVersion.Major}.{unityVersion.Minor}.{unityVersion.Build}");
+        var file = $"{unityVersion.Major}.{unityVersion.Minor}.{unityVersion.Build}.zip";
 
-        if (!string.IsNullOrEmpty(source))
-        {
-            Logger.LogMessage("Downloading unity base libraries");
-
-            Directory.CreateDirectory(UnityBaseLibsDirectory);
-            Directory.EnumerateFiles(UnityBaseLibsDirectory, "*.dll").Do(File.Delete);
-
+        var baseFolder = Directory.CreateDirectory(UnityBaseLibsDirectory);
+        baseFolder.EnumerateFiles("*.dll").Do(a=>a.Delete());
+        var target = baseFolder.GetFiles(file).FirstOrDefault();
+        if (target != null) {
+            using var fStream = target.OpenRead();
+            using var zipArchive = new ZipArchive(fStream, ZipArchiveMode.Read);
+            zipArchive.ExtractToDirectory(UnityBaseLibsDirectory);
+        } else {
+            var source = UnityBaseLibrariesSource.Value + file;
+            Logger.LogMessage($"Downloading unity base libraries {source}");
             using var httpClient = new HttpClient();
             using var zipStream = httpClient.GetStreamAsync(source).GetAwaiter().GetResult();
-            using var zipArchive = new ZipArchive(zipStream, ZipArchiveMode.Read);
-
             Logger.LogMessage("Extracting downloaded unity base libraries");
+            using var zipArchive = new ZipArchive(zipStream, ZipArchiveMode.Read);
             zipArchive.ExtractToDirectory(UnityBaseLibsDirectory);
         }
     }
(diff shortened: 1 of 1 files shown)


Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
```diff
@@ -282,25 +283,24 @@ private static void GenerateInteropAssemblies()
...
+
+        var baseFolder = Directory.CreateDirectory(UnityBaseLibsDirectory);
+        baseFolder.EnumerateFiles("*.dll").Do(a=>a.Delete());
+        var target = baseFolder.GetFiles(file).FirstOrDefault();
+        if (target != null) {
+            using var fStream = target.OpenRead();
+            using var zipArchive = new ZipArchive(fStream, ZipArchiveMode.Read);
+            zipArchive.ExtractToDirectory(UnityBaseLibsDirectory);
+        } else {
+            var source = UnityBaseLibrariesSource.Value.Replace("{VERSION}.zip", file);
```
ManlyMarco, line 298: Requiring the source setting to always add a zip at the end of `{VERSION}.zip` is a breaking change.

Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
```diff
@@ -282,25 +283,24 @@ private static void GenerateInteropAssemblies()
...
-
+        var file = $"{unityVersion.Major}.{unityVersion.Minor}.{unityVersion.Build}.zip";
+
+        var baseFolder = Directory.CreateDirectory(UnityBaseLibsDirectory);
+        baseFolder.EnumerateFiles("*.dll").Do(a=>a.Delete());
+        var target = baseFolder.GetFiles(file).FirstOrDefault();
+        if (target != null) {
+            using var fStream = target.OpenRead();
+            using var zipArchive = new ZipArchive(fStream, ZipArchiveMode.Read);
+            zipArchive.ExtractToDirectory(UnityBaseLibsDirectory);
```
ManlyMarco, line 296: This could be deduplicated with the same code below.

ManlyMarco (changes requested):

Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
lampuiho: well, System.IO.Compression.ZipArchive must be always a zip file but I'll revert this change. It just makes it easier to get the local file

Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
lampuiho: I'll end up with more lines that way because ZipArchive needs to be declared outside first. Then I need to add extra line to close the archive outside because I no longer can use "using". Using here makes this variable scoped so I can't extract it outside.

Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
ManlyMarco: It could be `{VERSION}_blah.zip` etc.

Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
ManlyMarco: Fair enough.

Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
```diff
@@ -282,25 +283,26 @@ private static void GenerateInteropAssemblies()
...
+        var version = $"{unityVersion.Major}.{unityVersion.Minor}.{unityVersion.Build}";
+        var source = UnityBaseLibrariesSource.Value.Replace("{VERSION}", version);
+        var uri = new Uri(source);
+        string file = Path.GetFileName(uri.AbsolutePath);
+
+        var baseFolder = Directory.CreateDirectory(UnityBaseLibsDirectory);
+        baseFolder.EnumerateFiles("*.dll").Do(a=>a.Delete());
+        var target = baseFolder.GetFiles(file).FirstOrDefault();
+        if (target != null) {
+            using var fStream = target.OpenRead();
```
ManlyMarco, line 297:
```suggestion
        if (target != null) {
            Logger.LogMessage($"Reading unity base libraries from file {source}");
            using var fStream = target.OpenRead();
```

Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
```diff
@@ -282,25 +283,26 @@ private static void GenerateInteropAssemblies()
...
-            Directory.CreateDirectory(UnityBaseLibsDirectory);
-            Directory.EnumerateFiles(UnityBaseLibsDirectory, "*.dll").Do(File.Delete);
-
+        var version = $"{unityVersion.Major}.{unityVersion.Minor}.{unityVersion.Build}";
+        var source = UnityBaseLibrariesSource.Value.Replace("{VERSION}", version);
+        var uri = new Uri(source);
+        string file = Path.GetFileName(uri.AbsolutePath);
+
+        var baseFolder = Directory.CreateDirectory(UnityBaseLibsDirectory);
+        baseFolder.EnumerateFiles("*.dll").Do(a=>a.Delete());
```
ManlyMarco, line 294: Before if UnityBaseLibrariesSource was empty nothing was done. Now it crashes when trying to create the Uri, which can definitely affect someone (also the old files are now always removed). Better to just keep the old behaviour.


```suggestion
        var source = UnityBaseLibrariesSource.Value.Replace("{VERSION}", version);
        if (string.IsNullOrEmpty(source)) return;

        var uri = new Uri(source);
        string file = Path.GetFileName(uri.AbsolutePath);

        var baseFolder = Directory.CreateDirectory(UnityBaseLibsDirectory);
        baseFolder.EnumerateFiles("*.dll").Do(a=>a.Delete());
```

ManlyMarco (changes requested):
After thinking about it, wouldn't it make sense to save the downloaded zip file to a known path (e.g. `(UnityBaseLibsDirectory)\UnityBaseLibraries.zip`) so that there is no need to rely on extracting the filename out of the URL which is not always guaranteed to work, and also to make it not necessary to redownload the libraries on a reload?

I guess that might cause issues if the game is updated to a newer Unity version, maybe `{UnityBaseLibsDirectory}\UnityBaseLibraries_{Version}.zip` then.

ManlyMarco (approved):

PR #1028, opened 2024-12-14
UnityVersion parsing and configuration

EvGem7:
## Description
It's essential for BepInEx to determine the unity version to work properly. This PR brings up a fix for the specific case when parsing of unity version goes wrong. Additionaly I've added an option to the config file, so the user can always set the version manually, if parsing silently fails again.

## Motivation and Context
Generally `BepInEx.Unity.Common.UnityInfo` works fine and determines the version correctly. But some games can have their files encrypted. For this case you would expect `AssetRipper.Primitives.UnityVersion::Parse` to fail. But actually it can "successfully" parse the encrypted bytes. The reason for this is that `UnityVersion::Parse` uses simple regexes under the hood and these regexes are not very strict. For example if you pass `THISISNOTAUNITYVERSION3!@#$%^&*(` to the `UnityVersion::Parse` it will successfully parse `3.0.0f1` version, since it has `3` somewhere within it.

This is exactly what happened to me. `UnityInfo` tried to parse the encrypted `globalgamemanagers` file and "successfully" parsed a unity version . This is how my file looks like:
image
UnityInfo tries to get a CString starting from 0x14 offset. Accidentally there is a `3` character in there.

I believe, there is no ideal fix to handle all possible cases automatically. But I tried to improve it a little. My fix is to skip version parsing, if there are non-printable characters in the CString.

Also, I'm not sure what was the motivation for adding the `GlobalMetadataPath` option to the config file, but this option really helped me to point BepIn to the decrypted file. Inspired by this option, I've added the UnityVersion field to the config file.

## How Has This Been Tested?
Smoke tests on the game with encypted files. Also tested on a "game" (actually just empty unity build) with usual non-encrypted files. Both of them run on the `2019.4.40f1` unity player version.

## Types of changes

## Checklist:

### DIFF
--- a/Runtimes/Unity/BepInEx.Unity.IL2CPP/Preloader.cs
+++ b/Runtimes/Unity/BepInEx.Unity.IL2CPP/Preloader.cs
@@ -27,7 +27,6 @@ public static class Preloader
         {
             HarmonyBackendFix.Initialize();
             ConsoleSetOutFix.Apply();
-            UnityInfo.Initialize(Paths.ExecutablePath, Paths.GameDataPath);
 
             ConsoleManager.Initialize(false, true);
 
@@ -43,6 +42,8 @@ public static class Preloader
             RedirectStdErrFix.Apply();
 
             ChainloaderLogHelper.PrintLogInfo(Log);
+            
+            UnityInfo.Initialize(Paths.ExecutablePath, Paths.GameDataPath);
 
             Logger.Log(LogLevel.Info, $"Running under Unity {UnityInfo.Version}");
             Logger.Log(LogLevel.Info, $"Runtime version: {Environment.Version}");
--- a/Runtimes/Unity/BepInEx.Unity.Common/BepInEx.Unity.Common.csproj
+++ b/Runtimes/Unity/BepInEx.Unity.Common/BepInEx.Unity.Common.csproj
@@ -6,6 +6,10 @@
     </PropertyGroup>
 
     <ItemGroup>
+        <ProjectReference Include="..\..\..\BepInEx.Core\BepInEx.Core.csproj"/>
+    </ItemGroup>
+
+    <ItemGroup>
         <PackageReference Include="AssetRipper.Primitives" Version="3.1.3" />
         <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" PrivateAssets="all"/>
         <PackageReference Include="MonoMod.Utils" Version="22.7.31.1"/>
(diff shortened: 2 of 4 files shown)


ManlyMarco (changes requested):

Runtimes/Unity/BepInEx.Unity.IL2CPP/Preloader.cs
```diff
@@ -43,6 +42,8 @@ public static void Run()
             RedirectStdErrFix.Apply();
 
             ChainloaderLogHelper.PrintLogInfo(Log);
+            
+            UnityInfo.Initialize(Paths.ExecutablePath, Paths.GameDataPath);
```
ManlyMarco, line 47: Why is delaying the initialization necessary?

EvGem7: To enable the error message be logged. Until the console is initialized the error log is useless.

ManlyMarco (approved):

js6pak:
@EvGem7 Assuming your game with encrypted assets is il2cpp, can you tell me what the following code prints out?
```cs
unsafe
{
    var get_dataPath = (delegate*unmanaged<IntPtr>)Il2CppInterop.Runtime.IL2CPP.il2cpp_resolve_icall("UnityEngine.Application::get_dataPath");
    Logger.Log(LogLevel.Warning, $"get_dataPath = {new string(Il2CppInterop.Runtime.IL2CPP.il2cpp_string_chars(get_dataPath()))}");
    var get_unityVersion = (delegate*unmanaged<IntPtr>)Il2CppInterop.Runtime.IL2CPP.il2cpp_resolve_icall("UnityEngine.Application::get_unityVersion");
    Logger.Log(LogLevel.Warning, $"get_unityVersion = {new string(Il2CppInterop.Runtime.IL2CPP.il2cpp_string_chars(get_unityVersion()))}");
}
```
If it's mono, I can write up an equivalent for it as well

ds5678:
Why wasn't this reported directly on my repository?

https://www.nuget.org/packages/AssetRipper.Primitives/3.1.4

EvGem7:
@js6pak I'm not sure, if the assets are encrypted. I know that the `globalgamemanagers` and `global-metadata.dat` are encrypted, at least parts of these files. Don't know much about other files.

The output of your code:
```
[Warning:   BepInEx] get_dataPath = C:/Games/GF2Exilium/GF2 Game/GF2_Exilium_Data
[Warning:   BepInEx] get_unityVersion = 2019.4.40f1
```

Oh no, now you know what is this mysterious game. :o

EvGem7:
@ds5678 now I'm thinking about this and I have no idea why I didn't do that. Yeah, I should've at least create an issue. Sorry, maybe I was thinking like it was intended? I don't know.

PrinceJohn15 (approved):

PR #1123, opened 2025-06-28
feat: new il2cppinterop generator flag for prefixing

soqb:
- we didn't have support for this flag in the config, which makes dealing with duplicate dependencies very tricky.
- could in the future allow specifying the values of the optin/optout arrays
- my formatter went haywire in the manager file so let me know if you want those changes reverted.


ManlyMarco:
That's fine, but the PR doesn't actually do anything other than add the config setting. It would have to be a full, finished feature to get merged.

This seems like a very niche use case to me, you could just as well rename your dependency to avoid conflicts.

ManlyMarco:
Closing rejected PR.

PR #1147, opened 2025-08-26
AllowDerived classes of BepInPlugin

EIA485:
allow 
```cs
class x: BepInPlugin {
    x(string GUID, string Name, string Version, string newArg):base(GUID,Name,Version) { }
}
[x("real.guid", "MyPlugin", "0.0.1", "example new field")]
Plugin : basePlugin {}
```
## Description
When the preloader scanning assemblies use .Resolve() recursively on types in order to find if any are derived from BepInPlugin

## Motivation and Context
This is useful for when modding communities need some extra game-specific metadata associated with a plugin.
yes they could create a derived type of BasePlugin with some extra abstract getters but this is cleaner.

## How Has This Been Tested?
This has been tested in the context of [ResoniteModdig/BepisLoader](https://github.com/ResoniteModding/BepisLoader)
with [BepInExResoniteShim](https://github.com/ResoniteModding/BepInExResoniteShim) as the plugin creating a derived type of BepInPlugin
any of [my BepisLoader mods](https://github.com/EIA485?tab=repositories&q=Neos&type=source&language=&sort=) as the plugin using said derived type

## Types of changes

## Checklist:
- [ ] My code follows the code style of this project. - not sure. could not find a code style document

## Notes:
if plugins are using a class defined in another plugin they should use [BepInDependency] to ensure their Dependency is loaded first.

### DIFF
--- a/BepInEx.Core/Bootstrap/BaseChainloader.cs
+++ b/BepInEx.Core/Bootstrap/BaseChainloader.cs
@@ -16,7 +16,7 @@ public abstract class BaseChainloader<TPlugin>
 {
     protected static readonly string CurrentAssemblyName = Assembly.GetExecutingAssembly().GetName().Name;
     protected static readonly Version CurrentAssemblyVersion = Assembly.GetExecutingAssembly().GetName().Version;
-
+    private static readonly Dictionary<string, bool> RefCache = new();
     private static Regex allowedGuidRegex { get; } = new(@"^[a-zA-Z0-9\._\-]+$");
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


EIA485:
Hey im unclear on how the examples you linked are related to this pr?
Are you saying i should extend BepInPlugin by defining a partial BepInPlugin class in my code? as far as i can tell partial classes across assemblies dont exist.

ManlyMarco:
> Hey im unclear on how the examples you linked are related to this pr?

They are not. It's just another way to accomplish something similar just with a shared project rather than separate assemblies.

ManlyMarco:
After looking at this again for a while I think this should not be merged into v6 since the feature set is mostly frozen now. It also adds too much complexity around a critical path for a niche use case.

It may be considered as a potential feature for v7, but has the same issue of added complexity.

PR #1187, opened 2025-10-06
Fix(IL2CPP): Correct handling of IL2CPPInteropAssembliesPath

Sleepywalker69:
## Description

This change corrects how the `IL2CPPInteropAssembliesPath` setting is handled.  
Previously, it was always treated as a base directory and a hardcoded `/interop` subfolder was appended, which caused incorrect path resolution when users provided a complete or custom path.  

The updated logic now treats `IL2CPPInteropAssembliesPath` as the *final* path when customised. The `/interop` suffix is only appended when the configuration remains at its default value of `{BepInEx}`, ensuring backward compatibility while allowing full control over path structure.

## Motivation and Context
 

This fix resolves an issue where specifying a custom interop assemblies path (such as `{ProcessName}` or an absolute directory) resulted in redundant or nested paths like `MyGame/MyGame/interop`.  
The change improves flexibility and predictability in how IL2CPP interop assemblies are located, making configuration more intuitive for users.

## How Has This Been Tested?
- Verified correct behaviour with default `{BepInEx}` configuration.  
- Tested custom absolute paths and relative templates (e.g., `{ProcessName}`).  
- Confirmed interop assemblies load correctly from user-defined directories.  

## Screenshots (if appropriate)
N/A

## Types of changes

## Checklist

### DIFF
--- a/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
+++ b/Runtimes/Unity/BepInEx.Unity.IL2CPP/Il2CppInteropManager.cs
@@ -113,8 +113,8 @@ internal static partial class Il2CppInteropManager
 
     private static readonly ManualLogSource Logger = BepInEx.Logging.Logger.CreateLogSource("InteropManager");
     
-    private static string _il2cppInteropAssemblyPath;
     private static string _il2cppBasePath;
+    private static string _il2cppInteropAssemblyPath;
 
     private static bool initialized;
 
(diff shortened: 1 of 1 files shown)


ManlyMarco (commented):
This will cause UnityBaseLibsDirectory to point inside of the interop directory. It may cause issues in the future if more things need to be added outside of the interop folder or handling of the downloaded assemblies changes.

Sleepywalker69 (commented):
Thanks for the feedback, I’ve updated the implementation to restore the correct dependency direction between IL2CPPBasePath and IL2CPPInteropAssemblyPath.

IL2CPPBasePath now acts as the primary root directory again, and IL2CPPInteropAssemblyPath is derived from it, ensuring future compatibility with potential folder structure changes (such as additional directories alongside interop).

This still preserves the intended behaviour of treating user-defined paths as final and only appending /interop when the configuration remains at the default {BepInEx} value.

PR #1234, opened 2025-12-28
Fix wine/proton monomod being loaded before bepinex

xinitrcn1:
## Description

monomod sets platform.current, then bepinex tries to set it, but you can only set it once, so it throws

## Motivation and Context

normal bepinex 5 didn't work, due to a infinite hangup, so i just went and said "well lets try 6"; everything was fine except for this singular line of code which made monomod raise an exception (i guess loading order in wine was different or whatever)

## How Has This Been Tested?

opened the game
bepinex now works

## Types of changes

## Checklist:

### DIFF
--- a/BepInEx.Preloader.Core/PlatformUtils.cs
+++ b/BepInEx.Preloader.Core/PlatformUtils.cs
@@ -127,7 +127,11 @@ internal static class PlatformUtils
                 current |= Platform.ARM;
         }
 
-        PlatformHelper.Current = current;
+        try {
+            PlatformHelper.Current = current;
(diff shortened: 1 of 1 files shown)


ManlyMarco (changes requested):
At the very least this should log a warning. If this fails there may be other random issues down the line so the cause needs to be checked more closely.

xinitrcn1:
> At the very least this should log a warning. If this fails there may be other random issues down the line so the cause needs to be checked more closely.

this is before logging begins

ManlyMarco:
Superseded by #1254

PR #1235, opened 2025-12-28
v5 - Fix for unity 2022.3.62f3 not having get_graphicsDeviceID

xinitrcn1:
## Description

fixes error with missing method

## Motivation and Context

allows one to run bepinex5

## How Has This Been Tested?

yeah

## Types of changes

## Checklist:

### DIFF
--- a/BepInEx/Bootstrap/Chainloader.cs
+++ b/BepInEx/Bootstrap/Chainloader.cs
@@ -48,7 +48,11 @@ namespace BepInEx.Bootstrap
 				var prop = AccessTools.PropertyGetter(typeof(Application), "isBatchMode");
 				if (prop != null)
 					return (bool) prop.Invoke(null, null);
-				return SystemInfo.graphicsDeviceID == 0;
+				try {
+					return SystemInfo.graphicsDeviceID == 0;
+				} catch(Exception e) {
+					return false; //Unity 2022.3.62f3 for example doesn't have such method
+				}
 			}
 		}
 
--- a/BepInEx/Bootstrap/Chainloader.cs
+++ b/BepInEx/Bootstrap/Chainloader.cs
@@ -50,8 +50,18 @@ namespace BepInEx.Bootstrap
 					return (bool) prop.Invoke(null, null);
 				try {
 					return SystemInfo.graphicsDeviceID == 0;
-				} catch(Exception e) {
-					return false; //Unity 2022.3.62f3 for example doesn't have such method
+				}
+				catch (MissingMethodException)
+				{
+					return false; // Unity 2022.3.62f3 for example doesn't have such method
+				}
+				catch (MissingFieldException)
+				{
+					return false; // Unity 2022.3.62f3 for example doesn't have such field
+				}
+				catch (MissingMemberException)
+				{
+					return false; // Unity 2022.3.62f3 for example doesn't have such member
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


ManlyMarco (commented):
Is there no viable alternative in case graphicsDeviceID is missing? How is this handled on the master branch, maybe the code could be backported?

BepInEx/Bootstrap/Chainloader.cs
```diff
@@ -48,7 +48,11 @@ private static bool IsHeadless
 				var prop = AccessTools.PropertyGetter(typeof(Application), "isBatchMode");
 				if (prop != null)
 					return (bool) prop.Invoke(null, null);
-				return SystemInfo.graphicsDeviceID == 0;
+				try {
+					return SystemInfo.graphicsDeviceID == 0;
+				} catch(Exception e) {
+					return false; //Unity 2022.3.62f3 for example doesn't have such method
+				}
```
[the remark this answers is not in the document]
ManlyMarco: It should use AccessTools like the code above instead of catching the exception.

xinitrcn1:
> Is there no viable alternative in case graphicsDeviceID is missing? How is this handled on the master branch, maybe the code could be backported?

no, unity 2022 just has no graphicsDeviceID, the only relatively sane way is to check DeviceName or whatever but that requires checking strings which can turn into a fun rabbit hole in and of itself (just ask the mesa people)

ManlyMarco:
That's weird, Unity docs say this property exists in all versions, is your game compiled for console/android? https://docs.unity3d.com/6000.3/Documentation/ScriptReference/SystemInfo-graphicsDeviceID.html

BepInEx/Bootstrap/Chainloader.cs
```diff
@@ -48,7 +48,21 @@ private static bool IsHeadless
...
+					return false; // Unity 2022.3.62f3 for example doesn't have such method
+				}
+				catch (MissingFieldException)
+				{
+					return false; // Unity 2022.3.62f3 for example doesn't have such field
+				}
+				catch (MissingMemberException)
+				{
+					return false; // Unity 2022.3.62f3 for example doesn't have such member
+				}
```
ManlyMarco: What exactly is the existing code throwing? If the method/property is missing then I would expect this trycatch to not prevent the crash, since it would crash when parsing the method body rather than on execution. I suggest switching to using AccessTools.PropertyGetter in that case. This will not work if the property getter exists but throws when called!

```suggestion
				// Some Unity builds may miss this property, see 
				// https://docs.unity3d.com/2022.3/Documentation/ScriptReference/SystemInfo-graphicsDeviceID.html
				var prop2 = AccessTools.PropertyGetter(typeof(SystemInfo), "graphicsDeviceID");
				if (prop2 != null)
					return ((int) prop2.Invoke(null, null)) == 0;
				
				return false;
```

xinitrcn1:
> That's weird, Unity docs say this property exists in all versions, is your game compiled for console/android? https://docs.unity3d.com/6000.3/Documentation/ScriptReference/SystemInfo-graphicsDeviceID.html

the docs lied, i checked the DLLs (unityjit-win32) it's not there

BepInEx/Bootstrap/Chainloader.cs
xinitrcn1: > What exactly is the existing code throwing? If the method/property is missing then I would expect this trycatch to not prevent the crash, since it would crash when parsing the method body rather than on execution. I suggest switching to using AccessTools.PropertyGetter in that case. This will not work if the property getter exists but throws when called!

middingmethod

ManlyMarco (approved):

ManlyMarco (approved):

Issue #1188, opened 2025-10-07
Game crashes after "Runtime invoke patched" on Unity 6000.1.17f1 (Dofus)


michael-tech24:
### What happened?

The game (Dofus Unity) crashes silently after BepInEx successfully initializes and patches the runtime. BepInEx loads all assemblies and reaches "Runtime invoke patched" but then the game closes immediately without any error message.

## BepInEx Distribution

Bleeding Edge from BepisBuilds - Tested with multiple versions:
- BepInEx 6.0.0-be.738
- BepInEx 6.0.0-be.697

## Game Information

- **Game:** Dofus (Unity version)
- **Unity Version:** 6000.1.17f1
- **IL2CPP:** Yes
- **Platform:** Windows 10 64-bit
- **Game works normally:** Yes (without BepInEx)

## Steps Taken

1. Installed BepInEx_UnityIL2CPP_x64 (multiple versions tested)
2. Game launches, BepInEx initializes successfully
3. Interop assemblies generate correctly
4. All assemblies load without errors
5. Chainloader initializes successfully
6. Runtime invoke gets patched
7. **Game closes immediately after** with no error

## Attempted Fixes

- ✅ Tried forcing Unity library versions: 6000.0.38, 6000.1.16
- ✅ Disabled `UnityLogListening = false`
- ✅ Enabled debug logging (`LogLevels = Fatal, Error, Warning, Message, Info, Debug`)
- ✅ Set `UpdateInteropAssemblies = false`
- ✅ Deleted and regenerated interop assemblies multiple times
- ✅ Tried different BepInEx bleeding edge builds (be.697, be.738)
- ✅ Set `DumpDummyAssemblies = true`
- ✅ Set `InstantFlushing = true`
- ✅ Launching game directly (not through launcher)
- ❌ **None of these fixed the crash**

## Log Output

BepInEx successfully initializes but game crashes after runtime patching:

`[Info   : Preloader] 0 assemblies discovered
[Message:AssemblyPatcher] Executing 0 patch(es)
[Message:   BepInEx] Chainloader initialized
[Debug  : Preloader] Runtime invoke pointer: 0x7FFF978FC2D0
[Debug  :DobbyDetour] Preparing detour from 0x7FFF978FC2D0 to 0x7FFF41BF38E4
[Debug  :DobbyDetour] Prepared detour; Trampoline: 0x7FFF57900000
[...]

Issue #1079, opened 2025-03-07
the game crash run [Chainloader initialized], Unity 6000.0.38f1


fishercc:
### What happened?

[Message: Preloader] BepInEx 6.0.0-be.733 -XXX (2025/2/26 15:28:56)
[Message: Preloader] Built from commit 995f04991b2b5ace4ddef07e12c3e99d4b2766a6
[Info   :   BepInEx] System platform: Windows 10 64-bit
[Info   :   BepInEx] Process bitness: 64-bit (x64)
[Info   :   BepInEx] Running under Unity 6000.0.38f1
[Info   :   BepInEx] Runtime version: 6.0.7
[Info   :   BepInEx] Runtime information: .NET 6.0.7
[Message:InteropManager] Downloading unity base libraries https://unity.bepinex.dev/libraries/6000.0.38.zip
[Message:InteropManager] Extracting downloaded unity base libraries
[Message:InteropManager] Running Cpp2IL to generate dummy assemblies from C:\Program Files\XXX\XXX_Data/il2cpp_data/Metadata/global-metadata.dat
[Info   :    Cpp2IL] [Library] Initializing Metadata...
[Info   :    Cpp2IL] [Library] Using actual IL2CPP Metadata version 31
[Info   :    Cpp2IL] [Library] Initialized Metadata in 160ms
[Info   :    Cpp2IL] [Library] Using binary type Portable Executable (from LibCpp2IL)
[Info   :    Cpp2IL] [Library] Searching Binary for Required Data...
[Info   :    Cpp2IL] [Library] Got Binary codereg: 0x181EEF7C0, metareg: 0x18219A570 in 260ms.
[Info   :    Cpp2IL] [Library] Initializing Binary...
[Info   :    Cpp2IL] [Library] Initialized Binary in 72ms
[Info   :    Cpp2IL] [Library] Mapping pointers to Il2CppMethodDefinitions...
[Info   :    Cpp2IL] [Library] Processed 68684 OK (89ms)
[Info   :    Cpp2IL] [Program] Creating application model...
[Info   :    Cpp2IL] [Program] Application model created in 1461.1653ms
[Info   :InteropManager] Cpp2IL finished in 00:00:05.5923563
[Info   :InteropManager] Generating interop assemblies
[Info   :Il2CppInteropGen] Reading assemblies...
[Info   :Il2CppInteropGen] Done in 00:00:00.0651372
[Info   :Il2CppInteropGen] Reading unity assemblies...
[...]

PR #1338, opened 2026-06-06
Degrade instead of crashing when the IL2CPP interop assemblies are missing

HetCreep:
When the interop assemblies aren't present at chainloader start, the game crashes silently instead of running unmodded.

`OnInvokeMethod` uses interop types directly in its body -- `Application.CallLogCallback(...)` and `LogType`, both from UnityEngine.CoreModule. That call is inside a try/catch, but it can't catch this: the JIT compiles the whole method before any of it runs, so when the interop assemblies can't be resolved the method throws `FileNotFoundException` at JIT time, before the try is entered. The exception unwinds into the native runtime-invoke detour and the process dies with nothing logged.

This moves the interop call into its own `SetupUnityLogging()` and calls it inside the try. `OnInvokeMethod` no longer references interop, so it JITs without the assemblies; `SetupUnityLogging()` is only JIT-compiled when called (inside the try), so a missing-interop failure is caught and the loader logs it and continues -- the game runs unmodded. I also set `unhook = true` at the top of the try so the detour fires exactly once even if setup throws.

Repro: an IL2CPP game on Unity 6000.0.58 with the interop folder empty. Before, the game CTDs silently right after "Runtime invoke patched". After, it logs `Unable to execute IL2CPP chainloader -- continuing unmodded` and launches. Interop can be missing for ordinary reasons too -- generation failing on an unsupported metadata version, or an incomplete interop folder.

This might also be behind the silent-crash reports in #1188 and #1079 (same OnInvokeMethod path, different Unity builds) -- worth a look from anyone who can still repro those.

### DIFF
--- a/Runtimes/Unity/BepInEx.Unity.IL2CPP/IL2CPPChainloader.cs
+++ b/Runtimes/Unity/BepInEx.Unity.IL2CPP/IL2CPPChainloader.cs
@@ -86,9 +86,7 @@ public class IL2CPPChainloader : BaseChainloader<BasePlugin>
                 // Unhook up front so the detour fires once even if setup below throws.
                 unhook = true;
 
-                // Isolated into its own method so OnInvokeMethod holds no interop type reference of its own: when the
-                // interop assemblies are missing, only this call fails to JIT -- inside the try, so it is caught --
-                // instead of OnInvokeMethod failing to JIT, which throws before the try is entered and crashes the game.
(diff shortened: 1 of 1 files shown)


ManlyMarco (changes requested):
Good fix but the comments are needlessly verbose. Condense them down to the bare minimum and change the log message to say something like "Unable to execute IL2CPP chainloader, no plugins will be loaded".

ManlyMarco (approved):

PR #1340, opened 2026-06-06
Do not abort startup when the config file cannot be written

### LINKED ISSUE #1339: Read-only BepInEx.cfg causes "Failed to start BepInEx" instead of degrading
## Summary

If `BepInEx/config/BepInEx.cfg` is read-only, BepInEx fails to start with a "Failed to start BepInEx" dialog and the game does not launch. A config file that cannot be written should degrade (keep the in-memory configuration and continue), not block startup.

## Repro

1. Install BepInEx (hit on IL2CPP, but the config path is runtime-agnostic).
2. Make `BepInEx/config/BepInEx.cfg` read-only (`attrib +R`, a read-only install directory, or an antivirus lock).
3. Launch the game.

Result: a "Failed to start BepInEx" dialog; the game does not start; **no `LogOutput.log` is written** -- the failure happens during config binding, before disk logging is initialized.

## Cause

`ConfigFile.Save()` opens the file for writing (`new StreamWriter(ConfigFilePath, false, ...)`) with no exception handling, and is invoked automatically via `SaveOnConfigSet` (default `true`) while settings are bound during preloader startup. On a read-only file the `StreamWriter` constructor throws `UnauthorizedAccessException`, which propagates out of the early startup path (before logging) and aborts BepInEx.

## Expected

An unwritable config file (read-only install directory, file locked by another process / antivirus) should not abort startup. The automatic saves should log a warning and continue with the in-memory configuration.

Environment: BepInEx 6 (master), Windows, Unity IL2CPP. Reproduced on a real game install with a read-only config.

HetCreep:
Fixes #1339

`ConfigFile.Save()` opens the file for writing with no exception handling and is invoked automatically via `SaveOnConfigSet` (default `true`) while settings are bound during startup, before logging is initialized. A read-only or otherwise unwritable config file -- a read-only install directory, or a file locked by another process / antivirus -- therefore throws `UnauthorizedAccessException` out of the early startup path and shows the "Failed to start BepInEx" dialog (with no `LogOutput.log`, since logging isn't up yet).

This routes the automatic saves (on init, on bind, on change) through a `TrySave` helper that logs a warning and keeps the in-memory configuration instead of propagating `IOException` / `UnauthorizedAccessException`. Explicit `Save()` calls are unchanged, so a plugin that calls `Save()` directly still sees the exception.

Repro: make `BepInEx/config/BepInEx.cfg` read-only, then launch -- before this change BepInEx fails to start; after it, it logs a warning and continues with the in-memory config.

### DIFF
--- a/BepInEx.Core/Configuration/ConfigFile.cs
+++ b/BepInEx.Core/Configuration/ConfigFile.cs
@@ -34,7 +34,7 @@ public class ConfigFile : IDictionary<ConfigDefinition, ConfigEntryBase>
 
         if (File.Exists(ConfigFilePath))
             Reload();
-        else if (saveOnInit) Save();
+        else if (saveOnInit) TrySave();
     }
 
     public static ConfigFile CoreConfig { get; } = new(Paths.BepInExConfigPath, true);
@@ -352,6 +352,24 @@ public class ConfigFile : IDictionary<ConfigDefinition, ConfigEntryBase>
         }
     }
 
+    /// <summary>
+    ///     Writes the config to disk like <see cref="Save" />, but logs and swallows I/O failures (e.g. a read-only
+    ///     or locked config file) instead of throwing. Used for the automatic saves so an unwritable config degrades
+    ///     to the in-memory configuration rather than aborting startup.
+    /// </summary>
+    private void TrySave()
+    {
+        try
+        {
+            Save();
+        }
+        catch (Exception e) when (e is IOException or UnauthorizedAccessException)
+        {
+            Logger.Log(LogLevel.Warning,
+                $"Could not write config file {ConfigFilePath}: {e.Message}. Continuing with the in-memory configuration.");
+        }
+    }
+
     #endregion
 
     #region Wraps
@@ -452,7 +470,7 @@ public class ConfigFile : IDictionary<ConfigDefinition, ConfigEntryBase>
             }
 
             if (SaveOnConfigSet)
-                Save();
+                TrySave();
 
             return entry;
         }
@@ -573,7 +591,7 @@ public class ConfigFile : IDictionary<ConfigDefinition, ConfigEntryBase>
         if (changedEntryBase == null) throw new ArgumentNullException(nameof(changedEntryBase));
 
         if (SaveOnConfigSet)
-            Save();
+            TrySave();
 
         var settingChanged = SettingChanged;
         if (settingChanged == null) return;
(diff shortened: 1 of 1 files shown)


ManlyMarco (commented):
Not sure about this one. I think this is a fatal issue and should crash rather than be silently ignored but I'm open to arguments otherwise.

HetCreep:
Fair concern — here's why I went with degrade-and-warn, but happy to change it if you still prefer fatal.

The save isn't silent: `TrySave` logs a `Warning` with the file path and the underlying error before continuing.

The reason I avoided crashing here specifically: the first automatic save runs while `ConfigFile.CoreConfig` is being constructed — before the logging backend is up. When the file is read-only/locked, the `UnauthorizedAccessException` escapes that static init and BepInEx aborts with "Failed to start BepInEx" and no `LogOutput.log` at all, so the user just sees the game fail to launch with nothing to go on. (Repro: mark `BepInEx.cfg` read-only → hard crash, empty log dir.)

Also, only the automatic saves go through `TrySave` (init + the `SaveOnConfigSet` paths). An explicit `Save()` still throws as before, so code that deliberately persists can still observe the failure.

That said, you know the project's expectations best. If you'd rather keep it fatal, one option is to log an actionable error and rethrow — though with the pre-logging timing above, a throw there still produces no log, whereas warn-and-degrade gives both the diagnostic and a launchable game. Glad to go whichever way you prefer.

ManlyMarco:
> The reason I avoided crashing here specifically: the first automatic save runs while `ConfigFile.CoreConfig` is being constructed — before the logging backend is up. When the file is read-only/locked, the `UnauthorizedAccessException` escapes that static init and BepInEx aborts with "Failed to start BepInEx" and no `LogOutput.log` at all, so the user just sees the game fail to launch with nothing to go on. (Repro: mark `BepInEx.cfg` read-only → hard crash, empty log dir.)

This is a very good point, I can see the utility now.

The problem with only crashing on Save is that almost no plugin ever uses it, everything instead relies on automatic saving on setting set. Doing this would effectively catch all config save exceptions.

Would it be enough to only catch exceptions on creation? I don't think BepInEx sets any of the settings internally, so the crash should only ever happen on creation.

HetCreep:
Done in 4459278 — narrowed exactly as you suggested:

- `OnSettingChanged` (the `SaveOnConfigSet` path plugins hit when setting values) goes back to `Save()` and throws as before.
- Only the creation-time saves keep the fallback: the initial file write in the constructor, and the save after binding a new entry — the bind during `CoreConfig` static init is the no-log crash window from #1339.

Builds clean across all TFMs.

ManlyMarco (approved):

PR #1344, opened 2026-06-11
Tolerate SemVer build metadata in plugin versions

### LINKED ISSUE #708: Mod can't load if build metadata is in AssemblyInformationalVersion
### What happened?

Mod can't load if build metadata is in AssemblyInformationalVersion(for example, `9.0.3.2+486c2711fc01e6deda1805da8950b5a4a692e7af`)

### Steps to reproduce

1. Build the mod with "IncludeSourceRevisionInInformationalVersion"  set to `true`(MSBuild 17.8.3(VS 2022 v17.8) has this value set to `true` by default)
`<IncludeSourceRevisionInInformationalVersion>true</IncludeSourceRevisionInInformationalVersion>`
2. Run game

### BepInEx Distribution

Bleeding Edge from BepisBuilds

### Log outputs

```txt
[Message: Preloader] BepInEx 6.0.0-be.671 - Among Us
[Message: Preloader] Built from commit 9caf61dca07043beae57b0771f6a5283aa02436b
[Info   :   BepInEx] System platform: Windows 10 64-bit
[Info   :   BepInEx] Process bitness: 32-bit (x86)
[Info   :   BepInEx] Running under Unity 2020.3.45f1
[Info   :   BepInEx] Runtime version: 6.0.7
[Info   :   BepInEx] Runtime information: .NET 6.0.7
[Info   : Preloader] 0 patcher plugins loaded
[Info   : Preloader] 0 assemblies discovered
[Message:AssemblyPatcher] Executing 0 patch(es)
[Message:   BepInEx] Chainloader initialized
[Info   :Il2CppInterop] Registered mono type Il2CppInterop.Runtime.DelegateSupport+Il2CppToMonoDelegateReference in il2cpp domain
[Warning:   BepInEx] Skipping type [ExtremeRoles.ExtremeRolesPlugin] because its version is invalid.
[Info   :   BepInEx] Loading [UnityExplorer 4.10.0]
[Message:UnityExplorer] UnityExplorer 4.10.0 initializing...
[Message:UnityExplorer] [UniverseLib] UniverseLib 1.5.4 initializing...
[Info   :Il2CppInterop] Registered mono type UniverseLib.UniversalBehaviour in il2cpp domain
[Info   :Il2CppInterop] Registered mono type UniverseLib.Runtime.Il2Cpp.Il2CppManagedEnumerator in il2cpp domain
[Message:UnityExplorer] [UniverseLib] Finished UniverseLib initial setup.
[Info   :Il2CppInterop] Registered mono type UnityExplorer.ExplorerBehaviour in il2cpp domain
[Message:UnityExplorer] Initialized UnityCrashPrevention for: Canvas.get_renderingDisplaySize
[Message:UnityExplorer] [UniverseLib] Loaded Unhollowed modules in 2.223 seconds.
[Message:UnityExplorer] [UniverseLib] Setup deobfuscation cache in 0.23600002 seconds.
[Message:UnityExplorer] [UniverseLib] Initialized Legacy Input support
[Message:UnityExplorer] [UniverseLib] 	 IL2CPP method has no corresponding pointer, aborting patch of UnityEngine.Cursor.lockState
[Message:UnityExplorer] [UniverseLib] 	 IL2CPP method has no corresponding pointer, aborting patch of UnityEngine.AssetBundle.UnloadAllAssetBundles
[Info   :Il2CppInterop] Registered mono type UniverseLib.AssetBundle in il2cpp domain
[Message:UnityExplorer] [UniverseLib] Loaded modern bundle for Unity 2020.3.45f1
[Message:UnityExplorer] [UniverseLib] UniverseLib 1.5.4 initialized.
[Message:UnityExplorer] Creating UI...
[Message:UnityExplorer] UnityExplorer 4.10.0 (IL2CPP) initialized.
```

### Environment

```markdown
- OS: Windows 10
- BepInEx: BepInEx 6.0.0-be.671
- Game: Among Us
```

HetCreep:
### What

Plugins whose version carries a SemVer build-metadata suffix (e.g. `9.0.3.2+486c2711...`) now load instead of being silently skipped.

### Why

`IncludeSourceRevisionInInformationalVersion` defaults to `true` in recent MSBuild/VS, which appends `+<commit>` to the informational version. When that string reaches `TryParseLongVersion`:

1. `SemanticVersioning.Version.TryParse` rejects it — the fourth component (`9.0.3.2`) isn't valid SemVer — so it falls through to the `System.Version` path.
2. `new System.Version("9.0.3.2+486c...")` throws on the `+`.
3. The `catch` returns `null`, `metadata.Version` is null, and the plugin is dropped.

The fix strips the build-metadata suffix (everything from `+`) before the `System.Version` parse. Versions without a `+` are untouched, so existing behavior is unchanged.

The same `TryParseLongVersion` exists in two places — `BepInPlugin` (Core) and `PatcherPluginInfo` (Preloader) — so both get the fix.

Fixes #708

Built across net35/net46/net6.

### DIFF
--- a/BepInEx.Core/Contract/Attributes.cs
+++ b/BepInEx.Core/Contract/Attributes.cs
@@ -53,7 +53,9 @@ public class BepInPlugin : Attribute
         // no System.Version.TryParse() on .NET 3.5
         try
         {
-            var longVersion = new System.Version(version);
+            // System.Version can't parse a SemVer build-metadata suffix (e.g. 1.2.3.4+sha); drop it first.
+            var plus = version.IndexOf('+');
+            var longVersion = new System.Version(plus >= 0 ? version.Substring(0, plus) : version);
 
             return new Version(longVersion.Major, longVersion.Minor,
                                longVersion.Build != -1 ? longVersion.Build : 0);
--- a/BepInEx.Preloader.Core/Patching/Attributes.cs
+++ b/BepInEx.Preloader.Core/Patching/Attributes.cs
@@ -46,7 +46,9 @@ public class PatcherPluginInfoAttribute : Attribute
         // no System.Version.TryParse() on .NET 3.5
         try
         {
-            var longVersion = new System.Version(version);
+            // System.Version can't parse a SemVer build-metadata suffix (e.g. 1.2.3.4+sha); drop it first.
+            var plus = version.IndexOf('+');
+            var longVersion = new System.Version(plus >= 0 ? version.Substring(0, plus) : version);
 
             return new Version(longVersion.Major, longVersion.Minor,
                                longVersion.Build != -1 ? longVersion.Build : 0);


ManlyMarco:
Under normal circumstances BepInVersion should never be set to the assembly/file version, since it requires the version to be included in the attribute parameters.

The only way for the project version to be used in BepInPlugin is by code generation, which is out of scope of this repository. Bugs caused by buggy custom build setups should be fixed by fixing those build setups, not making behaviour changes to BepInEx in order to accommodate improper use.

From what I've gathered the bug reports seem to all originate from AmongUs modding where a custom `BepInAutoPlugin` attribute is used that generates the plugin metadata dynamically. In this case `BepInAutoPlugin` should be updated with this fix, especially since `BepInAutoPlugin` does not seem to use `BepInPlugin` constructor underneath, therefore the fix in this repository would have no effect and not actually fix their issue.

PR #1351, opened 2026-06-15
Inheritance Submission

Dusk-Flare:
## Description

This change is simple, i added the `virtual` inheritance modifier to the Log methods, and also extracted the `LogEvent?.Invoke` call into it's own protected virtual method, that way the event could be called by inheriting classes, allowing them to better customise their logs with custom text such as custom exception handling, and text formatting such as my personal preference, text margins
## Motivation and Context

This is a simple change to the inheritance modifier of methods, and adds a layer between logging and invoking the log action, to allow for customisability of the log without compromising the original systems

## How Has This Been Tested?

**It has not been tested.**
It wall in my full intent to test this before making this request, but after a couple days of nonstop trying to build and apply to my project, i have not been able to test this, it builds, and the inheritance seams correct and like it wouldn't affect anything, but it would be wrong of me to claim it has been tested, please take this into consideration

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My code follows the code style of this project.

### DIFF
--- a/BepInEx.Core/Logging/ManualLogSource.cs
+++ b/BepInEx.Core/Logging/ManualLogSource.cs
@@ -52,7 +52,7 @@ public class ManualLogSource : ILogSource
     //     Internal method to allow for inheritance classes to have their own invocation logic.
     // </summary>
     /// <param name="arguments">Arguments to invoke the log event with.</param>
-    protected virtual void OnLogEvent(LogEventArgs arguments) => LogEvent?.Invoke(this, arguments));
+    protected virtual void OnLogEvent(LogEventArgs arguments) => LogEvent?.Invoke(this, arguments);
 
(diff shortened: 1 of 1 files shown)


ManlyMarco (commented):
Currently to do what you're describing you're supposed to just implement ILogSource yourself.

I think adding virtual to the Log methods doesn't make sense, it should be enough to only add the protected virtual OnLogEvent.

Dusk-Flare:
> Currently to do what you're describing you're supposed to just implement ILogSource yourself.
> 
> I think adding virtual to the Log methods doesn't make sense, it should be enough to only add the protected virtual OnLogEvent.

I am unsure of how i managed to mis this interface when making these changes, so sorry for the inconvenience

PR #1357, opened 2026-06-20
Enumerate a config snapshot under the lock

### LINKED ISSUE #1356: ConfigFile.GetEnumerator ignores the documented thread-safety contract
### What

`ConfigFile`'s class summary states *"All public methods are thread-safe."* but `GetEnumerator()` does not honor that contract.

### Detail

`BepInEx.Core/Configuration/ConfigFile.cs` (master 5f39645):

```csharp
public IEnumerator<KeyValuePair<ConfigDefinition, ConfigEntryBase>> GetEnumerator() =>
    // We can't really do a read lock for this
    Entries.GetEnumerator();
```

Every other member guards `Entries` with `lock (_ioLock)`, but `GetEnumerator` returns the live dictionary enumerator unlocked. If another thread calls `Bind()` or `Reload()` (which mutate `Entries`) while a consumer is iterating `foreach (var kv in configFile)`, the iterator throws `InvalidOperationException: Collection was modified`.

### Fix

Enumerate a snapshot taken under the lock (the same approach the logging collections already use):

```csharp
lock (_ioLock)
    return Entries.ToList().GetEnumerator();
```

PR attached.

HetCreep:
### What

`GetEnumerator()` returned the live `Entries` dictionary enumerator with no lock, despite the class documenting all public methods as thread-safe — a concurrent `Bind`/`Reload` could invalidate it mid-iteration.

### Fix

Snapshot `Entries` under `_ioLock` before enumerating (matching the snapshot approach the logging collections already use).

Fixes #1356

Built across net35/net46/net6.

### DIFF
--- a/BepInEx.Core/Configuration/ConfigFile.cs
+++ b/BepInEx.Core/Configuration/ConfigFile.cs
@@ -92,9 +92,12 @@ public class ConfigFile : IDictionary<ConfigDefinition, ConfigEntryBase>
     public ConfigEntryBase this[string section, string key] => this[new ConfigDefinition(section, key)];
 
     /// <inheritdoc />
-    public IEnumerator<KeyValuePair<ConfigDefinition, ConfigEntryBase>> GetEnumerator() =>
-        // We can't really do a read lock for this
-        Entries.GetEnumerator();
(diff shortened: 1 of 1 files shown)


ManlyMarco (approved):
Originally this was not done for better performance, but looking at how ConfigFile is used in the wild this should have negligible effect.

PR #1360, opened 2026-06-21
Add Input Console with command support

HayashiUme:
## Description
Add an **Input Console** feature that allows users to type commands directly into the BepInEx console to filter log output in real time or others.
### Commands
| Command | Description |
|---------|-------------|
| `/help` | Show available commands |
| `/show` | Reset filter — show all log sources |
| `/show <name>` | Show only logs from the specified source (case-insensitive exact match) |
| `/show <guid>` | Show only logs from a plugin by its GUID (resolved via internal GUID→SourceName mapping) |
| `/sources` | List all registered log sources with their plugin GUIDs |
| `/clear` | Clear the console screen |

## Motivation and Context
Currently, BepInEx's console displays all log output from all sources simultaneously. When debugging a specific plugin or subsystem, users have no way to filter the console output without modifying config files and restarting the game.
This feature enables real-time filtering directly from the console, significantly improving the debugging workflow for both plugin developers and users.

 ## How Has This Been Tested?
- Built successfully against all target frameworks (net35, netstandard2.0, net40, netstandard2.1, net6.0, netcoreapp3.1)
- Manually verified on IL2CPP runtime (Unity game):
  - `/help`, `/show`, `/show <name>`, `/show <guid>`, `/sources`, `/clear` all function correctly
  - Log filtering works — after `/show BepInEx`, only BepInEx source logs appear
  - Input line remains fixed at bottom during rapid log output
  - Command history navigation with ↑/↓ preserves and restores in-progress input
  - Unknown source warning displays correctly
- Verified old `ConsoleLogListener` instances from preloaders are properly cleaned up
- No breaking changes to existing logging APIs or plugin interfaces

``> /show cn.karped1em.tonx`` then open EmergencyMeetingMenu 
EF8DDEFC18AC259BC4012B71DF99E7C7

``> /show (reset show log source)`` then open EmergencyMeetingMenu
A10E58850D287A1D40C85581911097CE
## Types of changes

## Checklist:


HayashiUme:
Following the bot's instructions, some issues were fixed.

ManlyMarco (commented):
I have mixed feelings on this one. I can see it being useful if the game is super spammy in the log, but the way filtering works means you have to know the exact plugin name to filter for before it actually gets logged.

The current commands are a bit barebones. Typing unknown commands should show the help (or the default prompt should have /help in it just grayed out). The `show` command should be `filter` instead. There is no need for / slash before the commands. There should also be a way to filter by a text/regex match.

When I tried it I expected the console history to be filtered, and once filter was off then all of the logs to be restored, but I understand why doing this is not easy or ideal.

I'll let this sit for a while and see what others think, and if I warm up to the idea.

HayashiUme:
> I have mixed feelings on this one. I can see it being useful if the game is super spammy in the log, but the way filtering works means you have to know the exact plugin name to filter for before it actually gets logged.
> 
> The current commands are a bit barebones. Typing unknown commands should show the help (or the default prompt should have /help in it just grayed out). The `show` command should be `filter` instead. There is no need for / slash before the commands. There should also be a way to filter by a text/regex match.
> 
> When I tried it I expected the console history to be filtered, and once filter was off then all of the logs to be restored, but I understand why doing this is not easy or ideal.
> 
> I'll let this sit for a while and see what others think, and if I warm up to the idea.

Thank you for your patience in reviewing this. I do agree with your perspective to some extent.
Entering commands does indeed take a little time. If the code to be filtered is written inside `Load()`, it would be quite rushed. If we were to add a window to control the log output source on top of that, it would be even more troublesome and stray further from our original intent.
The extra `/` can indeed be removed. Initially, I intended for the console to recognize only inputs starting with `/` to prevent accidental entries, but that backfired, leading to the current implementation.
Based on my own testing, all features actually align with my design philosophy. Perhaps changing the command to `show Unity -override` would achieve the result you’re looking for?
In any case, I’m convinced this feature is very useful. Maybe it shouldn’t be limited to this niche “show logs” function; perhaps in the future, it could be developed into a method for plugins to register freely.

ManlyMarco:
Closing this after discussion on discord since the consensus was that this is not the way we want to go.

PR #1389, opened 2026-08-27
Don't kill the chainloader when Unity log callbacks are stripped

ObjectInSpace:
## Description

`UnityLogSource`'s static constructor subscribes to `Application.logMessageReceived`, falling back to the legacy `Application.RegisterLogCallback` when that event is missing:

```csharp
var registerLogCallback =
    typeof(Application).GetMethod("RegisterLogCallback", BindingFlags.Public | BindingFlags.Static);
registerLogCallback.Invoke(null, new object[] { callback });
```

Unity''s managed code stripper can remove **both** members. `GetMethod` then returns `null`, and the unguarded `Invoke` throws a `NullReferenceException` out of the static constructor.

This change guards the null and logs a warning instead, so Unity log forwarding degrades on its own rather than taking the loader down.

## Motivation and Context

Because the static constructor runs from chainloader initialisation, that `NullReferenceException` propagates up through the entrypoint''s `.cctor` and **the chainloader never starts — no plugins load at all**.

On the 5.x line, where I originally hit this, the failure is close to undiagnosable:

- `LogOutput.log` ends cleanly at `Preloader finished`, with no error.
- `BepInEx/cache/` stays empty, so it looks like plugins were never scanned.
- The exception appears **only on stdout**, which most users never capture.
- The game itself runs completely normally.

Captured stack trace (Beacon Pines, Unity 2021.3.45f1, Mono, BepInEx 5.4.23.5):

```
NullReferenceException: Object reference not set to an instance of an object
  at BepInEx.Logging.UnityLogSource..cctor ()
Rethrow as TypeInitializationException: The type initializer for ''BepInEx.Logging.UnityLogSource'' threw an exception.
  at BepInEx.Bootstrap.Chainloader.Initialize (System.String gameExePath, System.Boolean startConsole, ...)
  at UnityEngine.MonoBehaviour..cctor ()
```

Confirmed against that game''s `UnityEngine.CoreModule.dll` with Cecil:

```
Application events: [''onBeforeRender'', ''focusChanged'', ''quitting'']
has logMessageReceived event: False
has RegisterLogCallback method: False
```

Setting `UnityLogListening = false` works around it, but only once you already know that is the cause.

## How Has This Been Tested?

- `BepInEx.Unity.Mono` builds clean for both target frameworks (`net35` and `netstandard2.0`).
- Reproduced and fixed on a real game: Beacon Pines (Unity 2021.3.45f1, Mono, x64), whose `Application` has neither `logMessageReceived` nor `RegisterLogCallback`.
- **Verified on `master`.** This is reachable there once #1391 is applied — without it, `InitializeLoggers` fails eight lines earlier on the also-stripped `Application.unityVersion`, which masks this crash. With #1391 but *not* this PR:

  ```
  [Fatal] Unable to complete chainloader init: The type initializer for 'BepInEx.Unity.Mono.Logging.UnityLogSource' threw an exception.
    at BepInEx.Unity.Mono.Bootstrap.UnityChainloader.InitializeLoggers () ... UnityChainloader.cs:170
  ```

  With both applied, the same game starts cleanly and a test plugin's `Awake` runs:

  ```
  [Warning:   BepInEx] Unity log forwarding is unavailable: neither Application.logMessageReceived nor Application.RegisterLogCallback exist in this build ...
  [Message:   BepInEx] Chainloader initialized
  [Info   :   BepInEx] Loading [Probe6 1.0.0]
  [Message:    Probe6] PROBE6 AWAKE
  ```

- Also independently reproduced on the 5.x line, where it silently prevented the chainloader from starting; backported in #1392.
## Types of changes

## Checklist:

### DIFF
--- a/Runtimes/Unity/BepInEx.Unity.Mono/Logging/UnityLogSource.cs
+++ b/Runtimes/Unity/BepInEx.Unity.Mono/Logging/UnityLogSource.cs
@@ -2,6 +2,7 @@
 using System.Reflection;
 using BepInEx.Logging;
 using UnityEngine;
+using Logger = BepInEx.Logging.Logger;
 
 namespace BepInEx.Unity.Mono.Logging;
 
@@ -61,6 +62,23 @@ public class UnityLogSource : ILogSource
         {
             var registerLogCallback =
                 typeof(Application).GetMethod("RegisterLogCallback", BindingFlags.Public | BindingFlags.Static);
+
+            // Unity's managed code stripper can remove both Application.logMessageReceived and the
+            // legacy Application.RegisterLogCallback. Invoking a null MethodInfo here throws a
+            // NullReferenceException out of this static constructor, which propagates through
+            // Chainloader.Initialize and silently prevents the chainloader (and therefore every
+            // plugin) from starting. Unity log forwarding is optional, so degrade instead of
+            // taking the whole loader down with us.
+            if (registerLogCallback == null)
+            {
+                Logger.Log(LogLevel.Warning,
+                           "Unity log forwarding is unavailable: neither " +
+                           "Application.logMessageReceived nor Application.RegisterLogCallback exist " +
+                           "in this build (they were most likely stripped). Unity log messages will " +
+                           "not appear in the BepInEx log.");
+                return;
+            }
+
             registerLogCallback.Invoke(null, new object[] { callback });
             //UnsubscribeAction = () => registerLogCallback.Invoke(null, new object[] { null });
         }


ObjectInSpace:
Follow-up: I built this branch's `Unity.Mono-win-x64` distribution and ran it against the affected game (Beacon Pines, Unity 2021.3.45f1, Mono) to verify end to end. **The result does not support my earlier "How Has This Been Tested?" claim, so I want to correct it.**

On `master`, this code path is never reached on that game. `UnityChainloader.InitializeLoggers` fails earlier:

```
[Fatal  :   BepInEx] Unable to complete chainloader init: Method not found: string UnityEngine.Application.get_unityVersion()
  at BepInEx.Unity.Mono.Bootstrap.UnityChainloader.InitializeLoggers () ... UnityChainloader.cs:130
  at BepInEx.Bootstrap.BaseChainloader`1[TPlugin].Initialize ... BaseChainloader.cs:150
```

`Application.unityVersion` is stripped in that build too, so `UnityInfo.SetRuntimeUnityVersion(UnityVersion)` on line 130 throws before `Logger.Sources.Add(new UnityLogSource())` on line 138 ever runs.

I confirmed this by building and running both unpatched `master` and this branch against the same game: **the logs are byte-identical**, and in both cases the plugin is reported as "Loading [...]" and "Chainloader startup complete" but its `Awake` never executes.

So, accurately:

- The bug this PR fixes is **real and reproducible on the 5.x line**, where it silently kills the chainloader (that is where I originally hit it).
- On `master` the same unguarded `Invoke` still exists and is still a latent crash for any build that strips `logMessageReceived`/`RegisterLogCallback` but *not* `unityVersion` — but I have **not** been able to trigger it here.
- This change is therefore **defensive hardening on `master`, verified only by compilation**, not an end-to-end verified fix.

Happy to keep it as-is, or to close it if you would rather not carry a guard that has no reproducer on this branch.

Separately, `InitializeLoggers` looks like it has the same class of problem on line 130 — an aggressively stripped `UnityEngine.Application` takes out chainloader init. I can open a separate issue with the repro if that is useful.

ObjectInSpace:
Update: I can now verify this fix end-to-end on `master`, which I previously could not.

The reason it looked untriggerable is that a **second** stripped-member failure fires eight lines earlier in the same method. I have fixed that separately in #1391 (`Application.unityVersion` at `UnityChainloader.cs:130`).

With #1391 applied but **not** this PR, execution reaches line 170 and hits exactly the crash this PR guards:

```
[Fatal] Unable to complete chainloader init: The type initializer for 'BepInEx.Unity.Mono.Logging.UnityLogSource' threw an exception.
  at (wrapper managed-to-native) System.Object.__icall_wrapper_mono_generic_class_init(intptr)
  at BepInEx.Unity.Mono.Bootstrap.UnityChainloader.InitializeLoggers () ... UnityChainloader.cs:170
```

With **both** applied, the same game starts cleanly:

```
[Warning:   BepInEx] Unity log forwarding is unavailable: neither Application.logMessageReceived nor Application.RegisterLogCallback exist in this build (they were most likely stripped). Unity log messages will not appear in the BepInEx log.
[Message:   BepInEx] Chainloader initialized
```

So this is **reproducible on `master` after all** — it was simply masked by the earlier failure. My previous comment saying otherwise was correct about what I had observed at the time, but the conclusion was wrong.

The two PRs are independent but complementary: #1391 gets initialisation past the version read, this one gets it past the log-source constructor. Either alone leaves the other failure in place.

ObjectInSpace:
Tidied the branch into a single commit per CONTRIBUTING (the second commit was a fixup of the first). No content change.

ManlyMarco (approved):

Issue #1390, opened 2026-08-27
Stripped Application.unityVersion aborts chainloader init, but startup is still reported as complete


ObjectInSpace:
### Description

On a Unity Mono game with aggressive managed stripping, `UnityChainloader.InitializeLoggers` throws and chainloader init never completes — but BepInEx still reports success and plugins are silently inert.

### Reproduction

Game: Beacon Pines (Unity 2021.3.45f1, Mono, x64). Built `Unity.Mono-win-x64` from `master` (6abdba4) and installed it with a single trivial test plugin.

```
[Fatal  :   BepInEx] Unable to complete chainloader init: Method not found: string UnityEngine.Application.get_unityVersion()
[Fatal  :   BepInEx]   at BepInEx.Unity.Mono.Bootstrap.UnityChainloader.InitializeLoggers () ... UnityChainloader.cs:130
  at BepInEx.Bootstrap.BaseChainloader`1[TPlugin].Initialize (System.String gameExePath) ... BaseChainloader.cs:150
  at BepInEx.Unity.Mono.Bootstrap.UnityChainloader.Initialize (System.String gameExePath) ... UnityChainloader.cs:111
[Info   :   BepInEx] 1 plugin to load
[Info   :   BepInEx] Loading [Probe6 1.0.0]
[Message:   BepInEx] Chainloader startup complete
```

`UnityChainloader.cs:130` is `UnityInfo.SetRuntimeUnityVersion(UnityVersion)`, which reads `Application.unityVersion`. That property is stripped in this build:

```
Application methods present: OpenURL, Quit, get_dataPath, get_isEditor, get_isPlaying,
                             get_persistentDataPath, get_platform, get_version, ...
get_unityVersion present: False
has logMessageReceived event: False
has RegisterLogCallback method: False
```

### Two problems

**1. Init fails on a stripped optional member.** The version is only used for an informational log line, but throwing there aborts the rest of `InitializeLoggers`, including `Logger.Sources.Add(new UnityLogSource())` on line 138.

**2. The failure is reported as success.** After the `Fatal`, BepInEx logs `Loading [Probe6 1.0.0]` and `Chainloader startup complete`. The plugin is never actually started — my probe writes a file from `Awake` and that file is never created. From a user''s point of view the loader claims to have worked while every plugin is inert, which is very hard to diagnose.

### Suggestion

Guard the runtime version read (it is informational), and/or do not log `Chainloader startup complete` when initialisation threw.

[...]

PR #1391, opened 2026-08-27
Survive stripped UnityEngine.Application members during chainloader init

ObjectInSpace:
## Description

Two related failures when `UnityEngine.Application` has been aggressively stripped:

1. `InitializeLoggers` reads `Application.unityVersion` (line 130). The managed code stripper can remove it, and the resulting `MissingMethodException` aborts the rest of chainloader initialisation — including `Logger.Sources.Add(new UnityLogSource())` eight lines later. The version is only used for an informational log line, so it is now read defensively.

2. `UnityChainloader.Initialize` caught every exception, logged it as `Fatal`, and returned normally. `StaticStart` then called `Execute()` on a chainloader that was never initialised, so BepInEx logged **`Chainloader startup complete` while every plugin silently never ran**. Init failures now propagate.

## Motivation and Context

Observed on Beacon Pines (Unity 2021.3.45f1, Mono, x64), whose `Application` is missing `unityVersion`, `logMessageReceived` and `RegisterLogCallback`.

Unpatched `master`:

```
[Fatal  :   BepInEx] Unable to complete chainloader init: Method not found: string UnityEngine.Application.get_unityVersion()
  at BepInEx.Unity.Mono.Bootstrap.UnityChainloader.InitializeLoggers () ... UnityChainloader.cs:130
[Info   :   BepInEx] 1 plugin to load
[Info   :   BepInEx] Loading [Probe6 1.0.0]
[Message:   BepInEx] Chainloader startup complete
```

Note the last three lines: the loader reports success after a fatal error. A test plugin that writes a file from `Awake` never wrote it.

Fixes #1390.

## How Has This Been Tested?

Built the `Unity.Mono-win-x64` distribution from this branch and ran it against that game.

- **Before:** init aborted at `UnityChainloader.cs:130`; `Chainloader initialized` never logged.
- **After:** init proceeds past the version read and reaches `Chainloader initialized`.

With this fix alone, execution then advances to the *next* stripped-member failure at line 170 (`UnityLogSource`'s type initializer) — which is what #1389 guards:

```
[Fatal] Unable to complete chainloader init: The type initializer for 'BepInEx.Unity.Mono.Logging.UnityLogSource' threw an exception.
  at ... UnityChainloader.InitializeLoggers () ... UnityChainloader.cs:170
```

Applying both this PR and #1389 gets that game to a clean startup:

```
[...]

PR #1392, opened 2026-08-27
v5: Don't kill the chainloader when Unity log callbacks are stripped

ObjectInSpace:
## Description

Backport of #1389 to `v5-lts`.

`UnityLogSource`'s static constructor subscribes to `Application.logMessageReceived`, falling back to the legacy `Application.RegisterLogCallback` when that event is missing:

```csharp
MethodInfo registerLogCallback = typeof(Application).GetMethod("RegisterLogCallback", BindingFlags.Public | BindingFlags.Static);
registerLogCallback.Invoke(null, new object[] { callback });
```

Unity''s managed code stripper can remove **both** members. `GetMethod` then returns `null`, and the unguarded `Invoke` throws a `NullReferenceException` out of the static constructor.

## Motivation and Context

Because the static constructor runs from `Chainloader.Initialize`, that exception propagates up through the entrypoint''s `.cctor` and **the chainloader never starts — no plugins load at all**.

The failure is close to undiagnosable in practice:

- `LogOutput.log` ends cleanly at `Preloader finished`, with no error.
- `BepInEx/cache/` stays empty, so it looks like plugins were never scanned.
- The exception appears **only on stdout**, which most users never capture.
- The game itself runs completely normally.

Captured on Beacon Pines (Unity 2021.3.45f1, Mono, x64) with BepInEx 5.4.23.5:

```
NullReferenceException: Object reference not set to an instance of an object
  at BepInEx.Logging.UnityLogSource..cctor ()
Rethrow as TypeInitializationException: The type initializer for ''BepInEx.Logging.UnityLogSource'' threw an exception.
  at BepInEx.Bootstrap.Chainloader.Initialize (System.String gameExePath, System.Boolean startConsole, ...)
  at UnityEngine.MonoBehaviour..cctor ()
```

That game''s `UnityEngine.Application` has neither member:

```
Application events: [''onBeforeRender'', ''focusChanged'', ''quitting'']
has logMessageReceived event: False
has RegisterLogCallback method: False
```

Setting `UnityLogListening = false` works around it, but only once you already know that is the cause.

This is filed separately from #1389 because the file has moved and been restyled on `master` (`Runtimes/Unity/BepInEx.Unity.Mono/Logging/`, file-scoped namespaces, spaces), so the same diff does not apply to both branches.

## How Has This Been Tested?

Verified against the **official 5.4.23.5 release**, not a from-source build: installed the published `BepInEx_win_x64_5.4.23.5.zip` into the affected game (Beacon Pines, Unity 2021.3.45f1, Mono, x64) and replaced only `BepInEx/core/BepInEx.dll` with one built from this branch. `UnityLogListening` was left at its default `true` throughout — the setting that triggers the crash.

**Before (stock release DLL):** the log stops dead at `Preloader finished`, and the exception appears only on stdout:

```
NullReferenceException: Object reference not set to an instance of an object
  at BepInEx.Logging.UnityLogSource..cctor ()
Rethrow as TypeInitializationException: The type initializer for 'BepInEx.Logging.UnityLogSource' threw an exception.
```

No chainloader, no plugins, nothing in `LogOutput.log` to indicate a problem.

**After (this branch''s DLL):**

```
[Message:   BepInEx] Preloader finished
[Warning:   BepInEx] Unity log forwarding is unavailable: neither Application.logMessageReceived nor Application.RegisterLogCallback exist in this build (they were most likely stripped). Unity log messages will not appear in the BepInEx log.
[Message:   BepInEx] Chainloader ready
[Message:   BepInEx] Chainloader started
[Info   :   BepInEx] Loading [Beacon Pines Access 0.1.0]
```

The plugin loads and runs, and the `UnityLogListening = false` workaround is no longer needed.

`BepInEx.csproj` builds clean on this branch.

### Note

This game additionally needs a non-default `[Preloader.Entrypoint]` before plugins are actually ticked — filed separately as #1393. That is independent of this fix; both are required to get a working plugin on this title.
## Types of changes

## Checklist:

### DIFF
--- a/BepInEx/Logging/UnityLogSource.cs
+++ b/BepInEx/Logging/UnityLogSource.cs
@@ -59,6 +59,20 @@ namespace BepInEx.Logging
 			else
 			{
 				MethodInfo registerLogCallback = typeof(Application).GetMethod("RegisterLogCallback", BindingFlags.Public | BindingFlags.Static);
+
+				// Unity's managed code stripper can remove both Application.logMessageReceived and the
+				// legacy Application.RegisterLogCallback. Invoking a null MethodInfo here throws out of
+				// this static constructor, which propagates through Chainloader.Initialize and silently
+				// prevents the chainloader (and therefore every plugin) from starting. Unity log
+				// forwarding is optional, so degrade instead of taking the whole loader down.
+				if (registerLogCallback == null)
+				{
+					Logger.LogWarning("Unity log forwarding is unavailable: neither Application.logMessageReceived " +
+					                  "nor Application.RegisterLogCallback exist in this build (they were most likely " +
+					                  "stripped). Unity log messages will not appear in the BepInEx log.");
+					return;
+				}
+
 				registerLogCallback.Invoke(null, new object[] { callback });
 				//UnsubscribeAction = () => registerLogCallback.Invoke(null, new object[] { null });
 			}


ObjectInSpace:
For reviewers: this is one of four small PRs from the same investigation into a game with an aggressively stripped Unity Mono runtime. They are deliberately kept separate, per CONTRIBUTING''s "remain focused in scope" — each is one file, one logical change, and each stands alone.

**master**
- #1389 — `UnityLogSource..cctor` null-guard (logging)
- #1391 — `Application.unityVersion` read + don''t report a failed init as success (bootstrap), fixes #1390

**v5-lts**
- #1392 — backport of #1389 (the file moved and was restyled between branches, so the diffs differ)
- #1394 — default entrypoint to `MonoBehaviour`, fixes #1393

Nothing here depends on anything else being merged, with one exception worth knowing: on `master`, #1389''s crash is only *reachable* once #1391 is applied, because the version read fails eight lines earlier and masks it. Either can merge alone.

ManlyMarco (approved):

PR #1394, opened 2026-08-27
v5: Default the entrypoint to MonoBehaviour

### LINKED ISSUE #1393: Default Application..cctor entrypoint can load plugins that are never ticked (Update never runs)
### Description

On some Unity Mono games, the default `[Preloader.Entrypoint]` of `Application` / `.cctor` loads plugins but **never ticks them**. `BaseUnityPlugin.Awake` and `OnEnable` run, `Chainloader startup complete` is logged, and everything looks healthy — but `Update`, `LateUpdate` and coroutines never fire, so any per-frame plugin logic is silently dead.

### Environment

- BepInEx **5.4.23.5** (official `BepInEx_win_x64_5.4.23.5.zip`)
- Beacon Pines, Unity 2021.3.45f1, Mono, x64
- (This game also needs an unstripped corlib override to get this far; see #1392.)

### Symptoms

The log gives no indication of a problem:

```
[Message:   BepInEx] Chainloader started
[Info   :   BepInEx] Loading [Beacon Pines Access 0.1.0]
[Info   :Beacon Pines Access] ... plugin Awake logging, Harmony patches applied ...
```

But a heartbeat in `Update()` never fires. I confirmed the component itself looks entirely healthy from inside `OnEnable`:

```
isActiveAndEnabled=True  go=BepInEx_Manager  goActive=True  hideFlags=None  timeScale=1
```

`Awake` and `OnEnable` run because Unity executes those synchronously during `AddComponent`. The GameObject simply never gets registered with the update loop — presumably because `Application..cctor` runs before Unity is ready to do that.

I also verified this is not a logging artefact: a `File.WriteAllText` from `Update()`, bypassing BepInEx's logger entirely, never produced a file. Rewriting the per-frame work as a `WaitForEndOfFrame` coroutine did not help either, since coroutines are driven by the same loop.

### Workaround

Setting the entrypoint to `MonoBehaviour` fixes it completely:

```ini
[Preloader.Entrypoint]
Assembly = UnityEngine.CoreModule.dll
Type = MonoBehaviour
Method = .cctor
```

With that change, on the same game and the same plugin build:

```
[Info   :Beacon Pines Access] Update() is running.
[Info   :Beacon Pines Access] LateUpdate() is running.
```

and the plugin works normally.

### Suggestion

`MonoBehaviour` is already what BepInEx 6 uses by default. Two options that would both help:

1. Make `MonoBehaviour` the default entrypoint on 5.x as well, or
2. Add this to the troubleshooting docs alongside the existing "chainloader never starts" entry.

The existing troubleshooting guidance covers *"the preloader finishes and the chainloader never starts"*, but this failure is different and arguably worse: the chainloader **does** start, plugins **do** load, and everything reports success while nothing actually runs. There is currently nothing to search for when that happens.

ObjectInSpace:
## Description

Changes the default `[Preloader.Entrypoint]` `Type` from `Application` to `MonoBehaviour` on 5.x. One-line change.

## Motivation and Context

Fixes #1393.

On some Unity Mono games, `Application..cctor` runs before Unity is ready to register the `BepInEx_Manager` GameObject with the update loop. The result is that plugins load and appear completely healthy — `Awake` and `OnEnable` run, Harmony patches apply, `Chainloader startup complete` is logged — but `Update`, `LateUpdate` and coroutines **never fire**, so any per-frame plugin logic is silently dead.

From inside `OnEnable` the component state looks entirely normal:

```
isActiveAndEnabled=True  go=BepInEx_Manager  goActive=True  hideFlags=None  timeScale=1
```

`Awake`/`OnEnable` still run because Unity executes those synchronously during `AddComponent`, which is what makes this so hard to diagnose — there is nothing in the log to search for.

`MonoBehaviour..cctor` runs late enough to avoid the problem, and is **already the default in BepInEx 6**, so this also aligns the two branches.

### Compatibility

Existing installs are unaffected. `ConfigFile` writes `BepInEx.cfg` on first run, so any existing installation already has an explicit `Type = Application` on disk and keeps it. Only fresh installs pick up the new default.

Confirmed in `ConfigFile.Bind`: a key already present in the file is restored from `OrphanedEntries` via `SetSerializedValue`, so the new default only applies when the key is absent.

However, this *is* a change of shipped behaviour for **new** installs, so I have ticked "breaking change" as well — see the caveat under testing.

### Relationship to the existing docs

The published troubleshooting guide already recommends `Type = MonoBehaviour` for Unity 2017 and newer (and `MonoBehaviour` or `Camera` for Unity 5 and older). It does not mention `Application` at all. This change simply makes the shipped default match that existing guidance, so no documentation update should be needed.

## How Has This Been Tested?

- `BepInEx.Preloader` builds clean on this branch.
- The value itself is verified in normal use, not just in theory: my install of the affected game (Beacon Pines, Unity 2021.3.45f1, Mono, x64) has been running on the official 5.4.23.5 release with `Type = MonoBehaviour` set manually, and a screen-reader plugin that depends entirely on `Update()` works correctly. With the shipped default of `Application`, the same plugin loads, logs a clean startup, and never ticks once.

  Confirmed with a heartbeat in `Update()`, and independently with a `File.WriteAllText` that bypasses BepInEx logging entirely, in case the logger itself were at fault. Before the change the file was never written; after it, `Update() is running` and the plugin behaves normally.

**Caveat worth a maintainer's judgement:** I have tested this on exactly one game. If any title depends on the earlier Application..cctor timing, a fresh install would newly break on it, and I have no evidence either way. That risk is the main reason to consider this carefully rather than take my word for it.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) — for **new** installs only; see Compatibility

## Checklist:

### DIFF
--- a/BepInEx.Preloader/Preloader.cs
+++ b/BepInEx.Preloader/Preloader.cs
@@ -289,7 +289,7 @@ namespace BepInEx.Preloader
 
 		private static readonly ConfigEntry<string> ConfigEntrypointType = ConfigFile.CoreConfig.Bind(
 			"Preloader.Entrypoint", "Type",
-			"Application",
+			"MonoBehaviour",
 			"The name of the type in the entrypoint assembly to search for the entrypoint method.");
 
 		private static readonly ConfigEntry<string> ConfigEntrypointMethod = ConfigFile.CoreConfig.Bind(


ObjectInSpace:
For reviewers: this is one of four small PRs from the same investigation into a game with an aggressively stripped Unity Mono runtime. They are deliberately kept separate, per CONTRIBUTING''s "remain focused in scope" — each is one file, one logical change, and each stands alone.

**master**
- #1389 — `UnityLogSource..cctor` null-guard (logging)
- #1391 — `Application.unityVersion` read + don''t report a failed init as success (bootstrap), fixes #1390

**v5-lts**
- #1392 — backport of #1389 (the file moved and was restyled between branches, so the diffs differ)
- #1394 — default entrypoint to `MonoBehaviour`, fixes #1393

Nothing here depends on anything else being merged, with one exception worth knowing: on `master`, #1389''s crash is only *reachable* once #1391 is applied, because the version read fails eight lines earlier and masks it. Either can merge alone.

toebeann:
Breaking change that is unnecessary for v5. Highly unlikely to get merged.

Some games will benefit, yes. Those games people can just edit the .cfg.

In your linked issue, you mention the alternative is updating the docs. Would prefer that over this.

However in the troubleshooting section under the heading [Unity 2017 and newer](https://docs.bepinex.dev/articles/user_guide/troubleshooting.html#unity-2017-and-newer), this is already covered.

In other words, no change necessary.

ManlyMarco:
What toebeann said, this is a breaking change that isn't really necessary at this point for v5.

PR #1396, opened 2026-09-09
Support Version suffixes

### LINKED ISSUE #678: semver prerelease suffix not supported
### What happened?

Using a semantic version including a prerelease suffix; `1.1.0-beta.2`.

I'm using PluginInfoProps from the BepInEx Visual Studio template in order to generate the version string, and have put the above string as the project's display version. Both Visual Studio's description of the field and the semver specification explicitly supports prerelease suffixes, and the BepInEx documentation states that the version must be formatted according to semver.

I don't know whether this would be considered an issue with BepInEx, PluginInfoProps, or neither.

### BepInEx Distribution

Stable from GitHub

### Log outputs

```txt
[Warning:   BepInEx] Skipping type [MyPlugin] because its version is invalid.
```

### Environment

```markdown
- OS: Windows 10
- BepInEx: 5.4.22
```

bwoebi:
v6 supports proper semver versions, but v5 does not. Given that v6 is not released yet, and we'd like to officially support semver versions (e.g. Hexium supports them - and declaring it equivalently in the BepInPlugin would be very useful), we have to sort-of hack it in, because, obviously, simply changing the type would be a compatibility break.

Hence, we simply add VersionExtra to BepInPlugin to allow that information to be stored at all and avoid "because its version is invalid" errors.

Just stripping the suffixes away would be theoretically sufficient to avoid the warning, but also prevents it from being used in e.g. Bepinex.ConfigManager for display as well in the log showing the loaded mods and their versions.

Fixes #678.

## Types of changes

## Checklist:


ManlyMarco:
Why is this required for v5? The API for v5 has been frozen for a long time now and this would be a pretty major change that isn't backwards compatible if plugins started depending on it. All external tooling would also need to be updated to support this change otherwise the version would be rejected or incorrectly trimmed.

bwoebi:
We'd love to upgrade to v6, but it's still not yet released :-(

The change is backwards compatible - a new property is not an ABI break.
The version will not be rejected, just trimmed basically, if your tooling doesn't read the VersionExtra. Which is ... acceptable to me.

But yes, a plugin actually referencing the VersionExtra field would need to access it via reflection or declare itself incompatible with older BepInEx versions. But that's what minor version releases are for, to my understanding of semver.

arrowmaster:
> We'd love to upgrade to v6, but it's still not yet released :-(

v6 is suitable not for games using Mono. Even if v6 was released you would not be using it.

bwoebi:
@arrowmaster Oh yes, correct. So yeah, we're stuck with v5 and would still love to see that functionality.

ManlyMarco:
Personally I'm against the change, the versioning worked the same for entirety of v5 lifecycle and changing it at the very end just seems like unnecessarily inviting trouble. There's 4 digits in a version number, I refuse to believe it's not enough to convey version changes. Things like beta or RC can still be done with numbering.

If you really want this change then you can fork BepInEx and it would effectively have the same effect as far as plugin compatibility goes - plugins made for your version would not be compatible unless specifically designed in a way that allows running in old versions of BiE.

ManlyMarco:
Closing since consensus is to not merge.

PR #1399, opened 2026-09-12
v5: Combine dist releases and reduce to only arch zips

arrowmaster:
Simply future v5 releases by only having an x64 and x86 version that include UnityDoorstop for all platforms.


arrowmaster:
I will rebase after my other PR is merged so the CI can be tested

arrowmaster:
After discussion in Discord, I'm going to try and reword this to keep the existing release zips and include an additional \_all\_ zip that includes both x86 and x64 UnityDoorstop for all operating systems.

Issue #1310, opened 2026-04-30
The current release for MacOS cannot load plugins for Silksong


IDontNeedThisNow:
### What happened?

I updated to the most recent build, and tried running it. It works, and launches the game, but loads no plugins and only generates the config folder, and no log. The previous release, despite not being universal, works perfectly fine, and I have no idea of what could be going wrong.

### Steps to reproduce

1. Download BepInEx_macos_universal_5.4.23.5.zip and extract
2. Xattr doorstop, chmod and set the .app directory in the shell script
3. Cd to the directory, and run the script
4. It runs silksong with the normal terminal output, but doesn't load plugins, and only generates the config folder

### BepInEx Distribution

Stable from GitHub

### Log outputs

```txt
As previously mentioned, no log is generated. Sorry about that.
```

### Environment

```markdown
- OS: MacOS, Apple M3
- BepInEx: 5.4.23.5, macos-universal
- Game: Hollow Knight Silksong
```

toebeann:
what's going wrong is that bepinex is not actually ready for arm64, only the doorstop binary is which is the only universal piece in the release. but bepinex itself does not actually work with arm64 despite the doorstop binary supporting it. it can only work x64 right now due to some of its other dependencies.

as such, for now you need to force it to run x64 mode via rosetta. e.g.
```sh
/usr/bin/arch -x86_64 /bin/bash '/path/to/run_bepinex.sh'
```
and for use in steam launch options:
```sh
/usr/bin/arch -x86_64 /bin/bash '/path/to/run_bepinex.sh' %command%
```

IDontNeedThisNow:
I tried that, and it just doesn't work.

toebeann:
Try using gib to install BepInEx: https://github.com/toebeann/gib

I haven't had a chance to test it on M-series yet myself, but I'm told it works fine with latest BepInEx 5 on Apple Silicon.

If not, it may be that for now you have to just use the previous release - the newest one won't do much for you anyway. We're still a ways away from proper Apple Silicon support.

IDontNeedThisNow:
Upon further inspection, using the files from the current release, but the shell script from the previous one loads everything fine.

toebeann:
> Upon further inspection, using the files from the current release, but the shell script from the previous one loads everything fine.

Ah, yes this is on my radar. I believe gib ensures the script works properly, but the one currently included has issues for M series.

[...]

Issue #1303, opened 2026-04-21
macOS Apple Silicon (arm64) support: MonoMod/Harmony crashes in preloader


sbrothers7:
### Describe your problem

BepInEx 5.x cannot run natively on macOS Apple Silicon (arm64). The doorstop injection and preloader loading can be made to work (by compiling libdoorstop.dylib from source targeting arm64e), but the preloader crashes immediately due to MonoMod.RuntimeDetour.DetourHelper.GetIdentifiable() returning null on arm64 Mono. This blocks all Harmony patching, including HarmonyInteropFix.Apply() in Preloader.Run(), which cannot be bypassed via config.

I have underlined key information in my wiki site:
https://sbrothers7.github.io/docs/macOS/ADOFAI%20Modding/Native/

### Proposed solution

Any of these would unblock native arm64:

- Update the bundled MonoMod to a version with arm64 support (v25.x has it via MonoMod/MonoMod#241)
- Make HarmonyInteropFix.Apply() gracefully handle arm64 failure instead of crashing, so plugins that don't use Harmony can still load
- Add macos-arm64 as a build target for BepInEx 6.x bleeding edge

cdobbyn:
I ran into this on 5.4.23.5 and wanted to add what I found.

#1288 stops the crash, but `DetourHelper.Runtime` is still null afterwards, so Harmony patching is still dead. `HarmonyInteropFix.Apply()` at the top of `Preloader.Run()` throws, and you get no plugins, no crash, and no `LogOutput.log`. Quieter than before, same result.

The cause is `mprotect`. MonoMod uses `DetourNativeMonoPosixPlatform` on Mono, and its `MakeWritable` asks for `PROT_READ|PROT_WRITE|PROT_EXEC`. Unity has `allow-jit` but not `allow-unsigned-executable-memory`, so on Apple Silicon that combination isn't allowed and you get `EACCES`. Mono's code sits in a `MAP_JIT` region, which won't take `mach_vm_write` either.

The null you saw comes from `DetourHelper.get_Runtime`. It sets `_RuntimeInit` before building the platform, so once that constructor throws, every later call returns null. The real error happens once, and everything after it is a `NullReferenceException` somewhere else entirely.

[...]

Issue #899, opened 2024-06-10
MacOS arm64 (Apple Silicon) Support - Valheim


allquixotic:
### Describe your problem

Valheim just released on MacOS today. [Source](https://www.reddit.com/r/valheim/comments/1dch2nf/valheim_is_now_on_mac/)

On Apple Silicon (arm64) hardware, the game launches an arm64 binary. It is still built with Mono as evidenced by the usual Mono directories in the app bundle.

It appears that the Mac builds of BepInEx are all for Intel x64 platform.

### Proposed solution

Please provide builds of BepInEx for MacOS arm64, or create a Universal Binary. If you create a UB, then all the .dylib files for BepInEx will internally have both an Intel x64 build and an Apple arm64 build in the same file. 

A Universal Binary will be easier for users, as they just have to download "the Mac version" and not worry about which architecture their game is running. Some Unity games only ship Intel x64 binaries, but others run natively on arm64 on Apple Silicon.

This game is getting significant exposure at [WWDC 2024](https://developer.apple.com/wwdc24/) and may appear in the Keynote. Mods with BepInEx are very popular, therefore you can expect to see many, many requests for this until it is implemented.

allquixotic:
This is probably dependent on https://github.com/NeighTools/UnityDoorstop/issues/61 .

allquixotic:
Making some progress on https://github.com/NeighTools/UnityDoorstop/issues/61 --

Not sure if there is still a problem with the plthook_osx.c in Doorstop, but if there isn't, then something is amiss with Harmony.

See my preloader log for BepInEx/core/BepInEx.Unity.Mono.Preloader.dll on bleeding edge (it doesn't work with stable BepInEx core DLLs either):

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> HarmonyLib.HarmonyException: IL Compile Error (unknown location) ---> System.NullReferenceException: Object reference not set to an instance of an object
  at MonoMod.RuntimeDetour.DetourHelper.GetIdentifiable (System.Reflection.MethodBase method) [0x00005] in <7e79d87a23084646a4f7750a6feaa4c0>:0 
  at MonoMod.RuntimeDetour.ILHook..ctor (System.Reflection.MethodBase from, MonoMod.Cil.ILContext+Manipulator manipulator, MonoMod.RuntimeDetour.ILHookConfig& config) [0x0001c] in <7e79d87a23084646a4f7750a6feaa4c0>:0 
[...]

PR #1402, opened 2026-09-18
v5: Fix plugin loading on native arm64 macOS

cdobbyn:
## Description

On Apple Silicon, BepInEx loads no plugins and writes no log when the game runs as arm64 instead of under Rosetta. There is no crash and no error.

MonoMod picks `DetourNativeMonoPosixPlatform` on Mono, whose `MakeWritable`/`MakeExecutable` call `mprotect` with `PROT_READ|PROT_WRITE|PROT_EXEC`. macOS answers `EACCES`: without `com.apple.security.cs.allow-unsigned-executable-memory`, no page may be writable and executable at once. Unity ships `allow-jit` instead, so Mono's code is in a `MAP_JIT` region, which also refuses `mach_vm_write` with `KERN_INVALID_ADDRESS`.

`AppleSiliconDetourFix` installs a detour platform that keeps `DetourNativeARMPlatform`'s AArch64 encoding and routes code writes through `doorstop_jit_memcpy`, which brackets its `memcpy` with `pthread_jit_write_protect_np`. That toggle cannot be done from managed code — it removes execute permission for the calling thread, which is itself running JIT-compiled code. Two alternatives were tried and measured: calling the toggle directly gives SIGBUS, and `mach_vm_write` gives `KERN_INVALID_ADDRESS` on `MAP_JIT` pages while working on ordinary heap memory in the same process.

`XTermFix` forced `DetourNativeX86Platform` on every non-Windows target. On Apple Silicon that encodes x86 branches into arm64 method bodies and reinstates the failing `mprotect`, so it is now skipped when the new fix applies.

`Preloader.Run` logs a warning if the fix cannot install, beside the existing deferred warnings.

## Motivation and Context

Why the failure is invisible today, which is the reason this is worth fixing rather than documenting: `DetourHelper.get_Runtime` sets `_RuntimeInit` **before** it constructs the platform, and returns `null` on every later call once set. The `EACCES` escapes `DetourRuntimeILPlatform`'s constructor exactly once, and from then on every Harmony patch fails with a `NullReferenceException` in `DetourHelper.GetIdentifiable` — three frames from the cause. HarmonyX swallows the first occurrence inside `Harmony..cctor`, and `Doorstop.Entrypoint.Start` writes the rest to `preloader_*.log` beside the game executable rather than to `LogOutput.log`.

The practical effect is that Apple Silicon users must run their games under Rosetta to use BepInEx at all.

**Related issues.** Not using closing keywords — whether these are closed by this is the maintainers' call.

- #1303 describes this exact root cause: *"the preloader crashes immediately due to `MonoMod.RuntimeDetour.DetourHelper.GetIdentifiable()` returning null on arm64 Mono. This blocks all Harmony patching, including `HarmonyInteropFix.Apply()` in `Preloader.Run()`"*. It was closed by #1288, which wraps `XTermFix` and `ConsoleSetOutFix` in `TryDo`. That stops the crash, but `DetourHelper.Runtime` is still null afterwards, so Harmony patching remains broken — the preloader now fails quietly instead of loudly. This PR addresses the part that remained. #1288 is also a prerequisite: without it the preloader dies before reaching this code.
- #899, native arm64 support on Apple Silicon, reported against Valheim, which is the game used to verify this.
- #1310 reports 5.4.23.5 macos-universal loading no plugins and writing no log on an M3, while the older non-universal build works. That is the signature of this bug: the universal binary runs native arm64, the older one ran under Rosetta. I cannot test Silksong, so this is a likely match rather than a confirmed one.

**Dependency:** requires `doorstop_jit_memcpy`, proposed in NeighTools/UnityDoorstop#114. Until a Doorstop release carries it, this fix throws at construction and logs the new warning; behaviour is then exactly as it is today. Happy to add the `DOORSTOP_VER` bump in `build.cake` to this PR once that lands, if you would prefer them together.

I did not ask first, as CONTRIBUTING requests for significant changes. The diagnosis and the patch happened in one sitting, and it seemed more useful to bring the measurements than a question. Entirely understand if you would rather discuss the approach before reviewing code.

## How Has This Been Tested?

Environment: Apple M4 Pro, macOS 15, Valheim (Unity 6000.0.75f1, Mono), BepInEx 5.4.23.5, UnityDoorstop 4.5.0 built from source with the linked change.

Both architectures launched through the same script, differing only in `ARCHPREFERENCE`, with `LogOutput.log` deleted before each run:

| Run | Process | Result |
|---|---|---|
| arm64 | `Code Type: ARM64` | `Preloader finished`, `Chainloader startup complete`, plugin loaded, its Harmony patches applied |
| x86_64 | `Code Type: X86-64 (translated)` | unchanged from before the patch |

Confirmed on two separate builds of the test plugin. Before the patch the arm64 run produced no `LogOutput.log` at all; after it, the log is complete and the plugin reports its patches applied.

Also verified that MonoMod's own self-test now passes on arm64 — `DetourHelper.Runtime` resolves to `DetourRuntimeMonoPlatform` instead of staying `null`, which is what `DetourRuntimeILPlatform`'s constructor checks.

n/a

## Types of changes

## Checklist:


cdobbyn:
Trimmed the comments in `AppleSiliconDetourFix` (31 lines of comment down to 12). Dropped the account of how the old failure presented, since that stops being useful once this merges, and kept only the two facts that stop someone reverting the design: `mprotect` and `mach_vm_write` are both refused on MAP_JIT memory, and `pthread_jit_write_protect_np` cannot be called from managed code.

Recording the dropped part here, because it explains why this was never reported as a bug rather than as "plugins silently do not load":

`DetourHelper.get_Runtime` sets `_RuntimeInit` **before** it constructs the platform, and returns `null` on every later call once that flag is set. So the EACCES escapes `DetourRuntimeILPlatform`'s constructor exactly once, and from then on every Harmony patch fails with a `NullReferenceException` in `DetourHelper.GetIdentifiable` — three frames from the cause. HarmonyX swallows the first occurrence inside `Harmony..cctor`, and `Doorstop.Entrypoint.Start` writes the rest to `preloader_*.log` beside the game executable rather than to `LogOutput.log`. The result is a run with no plugins, no crash, and nothing in the BepInEx log.

The `Preloader.Run` warning added here exists so that outcome is not possible again.

cdobbyn:
One thing worth flagging: `master` looks like it has the same defect.

I assumed BepInEx 6 was on the MonoMod 25.x line, where `MacOSSystem` handles MAP_JIT properly. It isn't. `master` pins `MonoMod.RuntimeDetour` 22.7.31.1, and at that tag `DetourNativeMonoPosixPlatform.MakeWritable` still asks for `PROT_READ|PROT_WRITE|PROT_EXEC`, and `DetourHelper.get_Runtime` still sets `_RuntimeInit` before constructing the platform. `Runtimes/Unity/BepInEx.Unity.Mono.Preloader/RuntimeFixes/XTermFix.cs:34` also still has the unconditional `DetourHelper.Native = new DetourNativeX86Platform()`, comment included.

So a Mono game running natively on Apple Silicon under BepInEx 6 should fail the same way. That's read off the source though, not measured, so treat it as a lead rather than a report.

I didn't target `master` because none of the files here exist there and the port would be a rewrite against a different layout, which would make this harder to review rather than easier. The 6 side looks small, a fix class under `RuntimeFixes/` plus the `XTermFix` guard, but I'd want to run it on a real Mono game before opening anything. Happy to do that if you'd like a companion PR.

ManlyMarco (approved):
Looks good to me. It shouldn't affect currently working systems.

